PR #98, opened 2020-05-05
Avoid duplicates in credentials and presentations.

faustow:


src/verifiable-credential.js
```diff
@@ -39,24 +40,26 @@ class VerifiableCredential {
...
   /**
-   * Add a context to this Credential's context array
+   * Add a context to this Credential's context array. Duplicates are omitted.
    * @param {string|object} context - Context to add to the credential context array
    * @returns {VerifiableCredential}
    */
   addContext(context) {
     ensureObjectWithKeyOrURI(context, '@context', 'context');
-    this.context.push(context);
+    this.context = getUniqueElementsFromArray([...this.context, context], JSON.stringify);
```
lovesh, line 49: This will iterate over the collection `this.context` each time a context is added and stringify each element to detect duplicates. An efficient alternative would be to keep the collection as a Set and convert back to array and destringify in `toJSON`.

src/verifiable-credential.js
```diff
@@ -39,24 +40,26 @@ class VerifiableCredential {
...
   /**
-   * Add a type to this Credential's type array
+   * Add a type to this Credential's type array. Duplicates are omitted.
    * @param {string} type - Type to add to the credential type array
    * @returns {VerifiableCredential}
    */
   addType(type) {
     ensureString(type);
-    this.type.push(type);
+    this.type = [...new Set([...this.type, type])];
```
lovesh, line 61: Similar comment as above. Rather then creating a Set and converting back to array each time an element is added, you can keep it as set and make it array in `toJSON`.

src/verifiable-presentation.js
```diff
@@ -39,24 +40,24 @@ class VerifiablePresentation {
...
   /**
-   * Add a context to this Presentation's context array
+   * Add a context to this Presentation's context array. Duplicates are omitted.
    * @param {string|object} context - Context to add to the presentation context array
    * @returns {VerifiablePresentation}
    */
   addContext(context) {
     ensureObjectWithKeyOrURI(context, '@context', 'context');
-    this.context.push(context);
+    this.context = getUniqueElementsFromArray([...this.context, context], JSON.stringify);
```
lovesh, line 49: Same comment as above.

src/verifiable-presentation.js
```diff
@@ -39,24 +40,24 @@ class VerifiablePresentation {
...
   /**
-   * Add a type to this Presentation's type array
+   * Add a type to this Presentation's type array. Duplicates are omitted.
    * @param {string} type - Type to add to the presentation type array
    * @returns {VerifiablePresentation}
    */
   addType(type) {
     ensureString(type);
-    this.type.push(type);
+    this.type = [...new Set([...this.type, type])];
```
lovesh, line 60: Same comment as above.

src/verifiable-presentation.js
```diff
@@ -8,6 +8,7 @@ import {
 } from './utils/type-helpers';
```
lovesh, line 8: The changes made in this file are same as the previous one? Can use abstract this behavior in separate functions like `addUniqueContext`, `addUniqueType` and then use them at both places

src/verifiable-credential.js
```diff
@@ -9,6 +9,7 @@ import {
   ensureURI,
```
lovesh, line 9: What about uniqueness of subject?

src/verifiable-credential.js
faustow: Thanks! 
I believe your approach is more efficient if `addContext` is called more often than `toJSON`. If the opposite applies, the doing the conversions in `addContext` should be more efficient. 
It's hard to foresee which scenario will take place so we have to guess:
my guess is that `toJSON` will be called more often than `addContext` because it is possible to go through the entire issuing process without needing to add a new context, whereas `toJSON` is already called internally to sign and to verify even with no user interaction. Hence keeping my proposed approach would execute the conversions less often and be more efficient.

src/verifiable-credential.js
faustow: Same as above.

src/verifiable-presentation.js
faustow: Same as above.

src/verifiable-presentation.js
faustow: Same as above.

src/verifiable-credential.js
faustow: Good catch! The issue wasn't about it but I can add it here.

src/verifiable-presentation.js
faustow: Thanks! Maybe a superclass would be even cleaner, feels like an overkill atm though.

src/verifiable-credential.js
lovesh: I see that the context `https://www.w3.org/2018/credentials/examples/v1` is almost always needed which is added using the call and a more appropriate type will usually be added. But its fine, we are not adding 10s of contexts or types so let it be.

src/verifiable-presentation.js
lovesh: I wasn't implying a superclass, an independent util function taking the mutable array would work. No big deal.

src/verifiable-credential.js
lovesh: If by ticket you mean clubhouse then the title of the ticket mentions subject as well.

PR #100, opened 2020-05-07
Ensure to url encode dids sent to universal resolver

SamHellawell:


SamHellawell:
did qualifiers need url encoding due to the `:` character, so not sure a test is required

PR #101, opened 2020-05-07
Add VerifiableCredential fromJSON, setId, setIssuer and setProof methods

SamHellawell:


SamHellawell:
discussed with @faustow a little, need to do some updates here:
- fromjson should take extra properties we dont account for and assign them to vc object for interop
- tojson will return those properties for serializtion
- support non-array subject in fromjson
- write tests

PR #105, opened 2020-05-13
Validate cred schema

lovesh:


src/modules/schema.js
```diff
@@ -123,6 +125,17 @@ export default class Schema {
    * @returns {Boolean}
    */
   static validateSchema(json) {
+    let jsonSchemaSpec;
+    if (json['$schema']) {
+      if (json['$schema'].startsWith('http://json-schema.org/draft-07/schema')) {
+        jsonSchemaSpec = JSONSchema07;
+      } else {
+        // TODO: Fetch json['$schema'] (using http client) and convert function to async
```
SamHellawell, line 133: was unaware this method required http calls, in this case it should be marked as async in impl spec too

src/modules/schema.js
```diff
@@ -13,6 +14,7 @@ export const BLOB_ID_MAX_BYTE_SIZE = 32;
 // implementer may choose to implement this as a dynamic config option settable with the `parameter_type!` macro
 export const BLOB_MAX_BYTE_SIZE = 1024;
 
+// TODO: Remove
 // Schema-validator spec
 const jsonSchemaSpec = {
```
SamHellawell, line 19: happy to remove this, was purely for testing w/o http because was unsure if request should be here or somewhere else (since method of validateSchema was never marked async), thought maybe it could pass the schema obj somehow

tests/unit/schema.test.js
```diff
@@ -124,7 +124,7 @@ const exampleCredential = {
    ],
    "issuanceDate": "2020-03-18T19:23:24Z",
    "credentialSchema": {    // this is the schema
-      "id": "schema:dock:5C78GCA",
+      "id": "blobmodule:dock:5C78GCA",
```
SamHellawell, line 127: is this correct? `blobmodule` not `blob`?

src/modules/schema.js
```diff
@@ -123,6 +125,17 @@ export default class Schema {
    * @returns {Boolean}
    */
   static validateSchema(json) {
+    let jsonSchemaSpec;
+    if (json['$schema']) {
+      if (json['$schema'].startsWith('http://json-schema.org/draft-07/schema')) {
```
SamHellawell, line 130: not really a fan of using startswith/if statements here. if we can ensure url doesnt need startswith, we can do like :

```
const jsonSchemas = {
  'http://json-schema.org/draft-07/schema': JSONSchema07
};

const schemaID = json['$schema'];
if (jsonSchemas[schemaID]) {
  jsonSchemaSpec = jsonSchemas[schemaID];
}
```

SamHellawell: oh also, `["$schema"]` should be written in dot notion - i.e `json.$schema`

src/modules/schema.js
```diff
@@ -123,6 +125,17 @@ export default class Schema {
...
+        jsonSchemaSpec = JSONSchema07;
+      } else {
+        // TODO: Fetch json['$schema'] (using http client) and convert function to async
+      }
+    } else {
+      return false;
+    }
+    // Fixme: Is the intention to throw error or return false on error
```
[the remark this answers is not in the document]
lovesh: Agree, asked this because there was `return true` at the end. Will remove it so that a valid schema means no error

src/modules/schema.js
lovesh: Yes

tests/unit/schema.test.js
lovesh: It should be a `blob`

src/modules/schema.js
lovesh: I will abstract this in a function elsewhere, did that to illustrate the point.

src/modules/schema.js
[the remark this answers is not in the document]
SamHellawell: i dont think returning undefined makes sense here, maybe just do like `return validate` ?

src/modules/schema.js
```diff
@@ -124,22 +100,13 @@ export default class Schema {
...
-      return false;
-    }
-    // Fixme: Is the intention to throw error or return false on error
-    const result = validate(json, jsonSchemaSpec, {
-      throwError: true
+  static async validateSchema(json) {
+    // Get the JSON schema spec to check against.
+    const jsonSchemaSpec = await this.getJSONSchemaSpec(json);
+
+    validate(json, jsonSchemaSpec, {
```
SamHellawell, line 107: `return validate(json, ...` here - just incase user wants to see result and think its better than returning undefined

src/modules/schema.js
```diff
@@ -154,11 +121,42 @@ export default class Schema {
...
   }
+
+  /**
+   * Gets the JSON schema spec from given JSON. Will either return the stored JSON schema or get
+   * it using HTTP or will throw error if cannot get.
+   * @param {object} json
+   * @returns {Promise<object>}
+   */
+  static async getJSONSchemaSpec(json) {
+    const schemaKey = '$schema';
```
SamHellawell, line 139: useless variable, see comment below

src/modules/schema.js
```diff
@@ -154,11 +121,42 @@ export default class Schema {
...
+    if (json[schemaKey]) {
```
SamHellawell, line 140: use json.$schema

SamHellawell: ideally like `const schema = json.$schema;` then replace all `json[schemaKey]` references below with `schema`

src/modules/schema.js
```diff
@@ -154,11 +121,42 @@ export default class Schema {
...
+      // The URL might be 'http://json-schema.org/draft-07/schema' or 'http://json-schema.org/draft-07/schema#'
+      // In that case, the schema is already stored in the SDK as this is the latest JSON schema spec
+      if (json[schemaKey] === 'http://json-schema.org/draft-07/schema' || json[schemaKey] === 'http://json-schema.org/draft-07/schema-') {
```
SamHellawell, line 143: is `http://json-schema.org/draft-07/schema-` valid? comment above mentions `http://json-schema.org/draft-07/schema#`

src/utils/vc.js
```diff
@@ -306,10 +306,7 @@ export function buildDockCredentialStatus(registryId) {
  */
 export function validateCredentialSchema(credential, schema) {
   // TODO: The id will not be part of schema. The spec mentioned that id will be popped off from subject
-  // Fixme: Is the intention to throw error or return false on error
-  const result = validate(credential.credentialSubject, schema.schema || schema, {
-    throwError: true
+  validate(credential.credentialSubject, schema.schema || schema, {
```
SamHellawell, line 309: `return validate`

src/modules/schema.js
lovesh: Good catch.

src/modules/schema.js
lovesh: I need `schemaKey` for error but have cached the result of `json[schemaKey]` in a variable

PR #109, opened 2020-05-18
Enable schema validation on credential verification.

faustow:
Solves https://dock-team.atlassian.net/browse/DCK-20


src/verifiable-credential.js
```diff
@@ -126,6 +127,7 @@ class VerifiableCredential {
    */
   setSchema(id, type) {
     ensureURI(id);
+    // TODO: rename from `credentialSchema` to `schema` to remove redundancy, since we're in the `credential` class
```
SamHellawell, line 130: `credentialSchema` is part of spec, imo better to use this naming for interop

faustow: The same applies to other sections, they're called `credentialXxxxx` after `toJSON` gets called, but without the `credential` prefix inside the class. Wdyt?

SamHellawell: i dislike it because it forces user to use tojson when passing this object to a method that takes credentialSubject for example

```
const schema = new Schema();
verify(schema.credentialSubject, schemaRef);
```

rather than

```
const schema = new Schema();
verify(schema.toJSON().credentialSubject, schemaRef);
```

faustow: `credentialSubject` is the only param in the `VerifiableCredential` class that breaks the non-redundancy rule, hence my comment of fixing it. The idea is simple: when using an instance of the `VerifiableCredential` class, the `credential` prefix is implied and not needed.
VerifiableCredentials' own `verify` should never try to use a param with the `credential`-prefix.

SamHellawell: i think you are missing my point entirely about passing objects around and not having to call tojson as a user to get a proper instance of VC data.

also you are incorrect, because there is `credentialSchema` field aswell as `credentialSubject`

tests/integration/schema.test.js
```diff
@@ -18,12 +18,9 @@ let pair;
 let publicKey;
 let dockDID;
 let keyDetail;
-let txDid;
-// eslint-disable-next-line no-unused-vars
-let resultDid;
-// eslint-disable-next-line no-unused-vars
-let didDoc;
 let blobId;
+// eslint-disable-next-line no-unused-vars
```
SamHellawell, line 22: shouldnt be a need to use eslint disable rule here

src/verifiable-credential.js
faustow: I see what you, however calling `toJSON` is only needed when operating outside of the context of a `VerifiableCredential` class instance. 

> ...incorrect, because there is `credentialSchema` field aswell as `credentialSubject`

.. both of which were added in this branch. This adds to my point that these changes break the pre-existing style in both `VerifiableCredential` and `VerifiablePresentation` classes ("avoiding the `credential` prefix inside of the class, and only using it outside").

faustow: Wdyt @bddap @lovesh ? maybe we can vote. Whatever way we choose we should fix the rest and make sure there are no inconsistencies in this style to provide a better user experience.

lovesh: We have to call `toJSON` in some places anyway as it converts context to `@context`. I would prefer to keep them `credentialSchema` and so on and do minimal work in `toJSON` as I had raised this point in an earlier PR with subject and other fields.

faustow: On one hand you mention that `toJSON` is needed in several places anyway since `@context` cannot be used ad a property name. Yet you'd prefer to have redundant property names like `credentialSchema` and reduce `toJSON` to a minimum? Not sure I understand the connection there. Since we can't get rid of using `toJSON` anyway, wouldn't it be better to leverage it and provide a non-redundant (better) internal api? That way you can use `credential.schema` instead of `credential.credentialSchema`.

PR #110, opened 2020-05-19
Schema [Jira: DCK-58]

SamHellawell:


src/modules/blob.js
```diff
@@ -0,0 +1,86 @@
...
+import { getSignatureFromKeyringPair, getStateChange } from '../utils/misc';
+import NoBlobError from '../utils/errors/no-blob-error';
+
+export const DockBlobMethod = 'dock';
+export const DockBlobQualifier = `blob:${DockBlobMethod}:`;
+export const DockBlobIdByteSize = 32;
+
+/**
+ * Create and return a fully qualified Dock Blob, i.e. "did:dock:<SS58 string>"
+ * @returns {string} - The Blob
```
lovesh, line 11: The Blob -> The Blob Id

src/modules/blob.js
```diff
@@ -0,0 +1,86 @@
...
+ */
+export function createNewDockBlobId() {
+  const hexId = randomAsHex(DockBlobIdByteSize);
+  return blobHexIdToQualified(hexId);
+}
+
+/**
+ * Return a fully qualified Dock Blob, i.e. "did:dock:<SS58 string>"
```
lovesh, line 19: qualified Dock Blob -> qualified Dock Blob Id

src/modules/blob.js
```diff
@@ -0,0 +1,86 @@
...
+    this.module = api.tx.blobStore;
+  }
+
+  /**
+   * Register a new Blob on the Dock Chain
+   * @param {object} blob - struct to store on chain
+   * @param {object} keyPair - Key pair to sign with
+   * @return {object} The extrinsic to sign and send.
+   */
+  new(blob, keyPair) {
```
lovesh, line 45: Same comment as i had made to revocation methods, it should be possible to pass in signature from outside (hardware wallet, HSM, etc)

src/modules/blob.js
```diff
@@ -0,0 +1,86 @@
...
+    return this.module.new(blob, signature.toJSON());
+  }
+
+  /**
+   *  Get blob with given id from the chain. Throws if the blob can't be found.
+   * @param {string} id - Can either be a full blob id like blob:dock:0x... or just the hex identifier
+   * @returns {Promise<Array>} - A 2-element array where the first is the author and the second is the blob contents.
+   */
+  async getBlob(id) {
+    id.replace(DockBlobQualifier, '');
```
lovesh, line 57: `replace` doesn't mutate given string but returns a new string. Secondly why not just `id = id.slice(DockBlobQualifier)`, its faster and does exactly what needs to be done. You don't need to search in the whole string. What if the qualifier was added to the hex id and not ss58? Have a look at the `getDocument` for did module.

src/modules/blob.js
```diff
@@ -0,0 +1,86 @@
...
+    const resp = await this.api.query.blobStore.blobs(id);
+    if (resp.isNone) {
+      throw new NoBlobError(id);
+    }
+    const tuple = resp.unwrap();
+    return [tuple[0], tuple[1]];
```
lovesh, line 63: What if return value is not of length 2, have a look at `getDetail` of did module

src/modules/schema.js
```diff
@@ -0,0 +1,240 @@
...
+import { isHexWithGivenByteSize } from '../utils/codec';
+import Signature from '../signatures/signature';
+
+import { DockBlobIdByteSize } from './blob';
+
+// Supported schemas
+import JSONSchema07 from '../utils/vc/schemas/schema-draft-07';
+
+// Blob qualifier
+export const BlobQualifier = 'blob:dock:';
```
lovesh, line 17: This was defined in blob module as well.

src/modules/schema.js
```diff
@@ -0,0 +1,240 @@
...
+
+// Size of the blob id in bytes
+export const BLOB_ID_MAX_BYTE_SIZE = 32;
```
lovesh, line 20: Defined in blob module as well

src/modules/schema.js
```diff
@@ -0,0 +1,240 @@
...
+
+// Maximum size of the blob in bytes
+// implementer may choose to implement this as a dynamic config option settable with the `parameter_type!` macro
+export const BLOB_MAX_BYTE_SIZE = 1024;
```
lovesh, line 24: This belongs in blob module

src/modules/schema.js
```diff
@@ -0,0 +1,240 @@
...
+
+/**
+ * Check if the given identifier is 32 byte hex
+ * @param {string} identifier - The identifier to check.
+ * @return {void} Throws exception if invalid identifier
+ */
+export function validateBlobDIDHexIdentifier(identifier) {
```
lovesh, line 31: The function name has DID but its not about a DID. And the logic is same as `validateDockDIDHexIdentifier`. Abstract the common part in a separate function to accept identifier and qualifier.

src/modules/schema.js
```diff
@@ -0,0 +1,240 @@
...
+  // Byte size of the Dock DID identifier, i.e. the `DockDIDQualifier` is not counted.
```
lovesh, line 32: Misleading comment.

src/modules/schema.js
```diff
@@ -0,0 +1,240 @@
...
+  if (!isHexWithGivenByteSize(identifier, BLOB_ID_MAX_BYTE_SIZE)) {
+    throw new Error(`DID identifier must be ${BLOB_ID_MAX_BYTE_SIZE} bytes`);
```
lovesh, line 34: Misleading error.

src/modules/schema.js
```diff
@@ -0,0 +1,240 @@
...
+  }
+}
+
+/**
+ * Gets the hexadecimal value of the given DID.
```
lovesh, line 39: The doc comments are wrong. Its not about a DID

src/modules/schema.js
```diff
@@ -0,0 +1,240 @@
...
+ * @param {string} did -  The DID can be passed as fully qualified DID like `blob:dock:<SS58 string>` or
+ * a 32 byte hex string
+ * @return {string} Returns the hexadecimal representation of the DID.
+ */
+export function getHexIdentifierFromBlobDID(did) {
```
lovesh, line 44: Its blob id not DID.

lovesh, line 44: The logic is copied from `getHexIdentifierFromDID`. Don't duplicate please. Abstract the logic to accept qualifier, size and error message

src/modules/schema.js
```diff
@@ -0,0 +1,240 @@
...
+    const ss58Did = did.slice(BlobQualifier.length);
+    try {
+      const hex = u8aToHex(decodeAddress(ss58Did));
+      // 2 characters for `0x` and 2*byte size of DID
+      if (hex.length !== (2 + 2 * BLOB_ID_MAX_BYTE_SIZE)) {
+        throw new Error('Unexpected byte size');
+      }
+      return hex;
+    } catch (e) {
+      throw new Error(`Invalid SS58 DID ${did}. ${e}`);
```
lovesh, line 56: Not DID

src/modules/schema.js
```diff
@@ -0,0 +1,240 @@
...
+    }
+  } else {
+    try {
+      // Check if hex and of correct size and return the hex value if successful.
+      validateBlobDIDHexIdentifier(did);
+      return did;
+    } catch (e) {
+      // Cannot parse as hex
+      throw new Error(`Invalid hexadecimal DID ${did}. ${e}`);
```
lovesh, line 65: Not DID

src/modules/schema.js
```diff
@@ -0,0 +1,240 @@
...
+    }
+  }
+}
+
+/**
+ * Generates a random schema ID
+ * @returns {string}
+ */
+export function createNewSchemaID() {
```
lovesh, line 74: Why not just call `createNewDockBlobId` inside?

src/modules/schema.js
```diff
@@ -0,0 +1,240 @@
...
+      ...rest
+    } = this;
+    return rest;
+  }
+
+  /**
+   * Serializes the schema to a blob object to send to the node
+   * @returns {object}
+   */
+  toBlob(id, did) {
```
lovesh, line 164: This function here seems weird. If you want to have a function that creates a blob from a JSON schema, have a detached one or have static in Blob like `fromJSONSchema`
Whats the point of passing id? This object will always have an id? Why allow it to be changed.

src/modules/schema.js
```diff
@@ -0,0 +1,240 @@
...
+    return {
+      id: id || randomAsHex(DockBlobIdByteSize),
+      blob: u8aToHex(u8aToU8a(JSON.stringify(this.toJSON()))),
```
lovesh, line 167: `u8aToHex(u8aToU8a` -> `stringToHex`

src/modules/schema.js
```diff
@@ -0,0 +1,240 @@
...
+   * @returns {Promise<object>}
+   */
+  static async getSchema(id, dockApi) {
+    const hexId = getHexIdentifierFromBlobDID(id);
+    const chainBlob = await dockApi.blob.getBlob(hexId);
+    const blobStr = u8aToString(chainBlob[1]);
+    try {
+      const schema = JSON.parse(blobStr);
+      schema.id = id;
+      schema.author = u8aToHex(chainBlob[0]);
```
lovesh, line 201: `blob:dock:` prefix

src/modules/schema.js
SamHellawell: `Whats the point of passing id? This object will always have an id? Why allow it to be changed.` ID of blob is unknown from the Schema object unless its generated externally

src/modules/schema.js
SamHellawell: method was added/used before blob module was integrated, could do that

src/modules/schema.js
lovesh: A `Schema` object will always have an id, if you don't pass in the constructor, it will be auto generated

SamHellawell: yeah not sure why i wrote it this way for the example, have resolved

src/modules/blob.js
faustow: Kept the keypair option and allowed signature to be sent too.

example/schema.js
```diff
@@ -0,0 +1,117 @@
...
+  const account = dock.keyring.addFromUri(TestAccountURI);
+  dock.setAccount(account);
+
+  // Generate aDID to be used as author
+  const dockDID = createNewDockDID();
+
+  console.log('Creating new DID', dockDID);
+
+  // Generate first key with this seed. The key type is Sr25519
+  const pair = dock.keyring.addFromUri(randomAsHex(32));
```
lovesh, line 35: Similar comment as another example, setup code should be kept separate

src/modules/blob.js
```diff
@@ -0,0 +1,111 @@
...
+ * @param {string} id -  The ID can be passed as fully qualified ID like `blob:dock:<SS58 string>` or
+ * a 32 byte hex string
+ * @return {string} Returns the hexadecimal representation of the ID.
+ */
+export function getHexIdentifierFromBlobID(id) {
+  return getHexIdentifier(id, DockBlobQualifier, validateBlobIDHexIdentifier, DockBlobIdByteSize);
+}
+
+/**
+ * Create and return a fully qualified Dock Blob id, i.e. "did:dock:<SS58 string>"
```
lovesh, line 36: did:dock -> blob:dock

src/modules/blob.js
```diff
@@ -0,0 +1,111 @@
...
+ * @returns {string} - The Blob id
+ */
+export function createNewDockBlobId() {
+  const hexId = randomAsHex(DockBlobIdByteSize);
+  return blobHexIdToQualified(hexId);
+}
+
+/**
+ * Return a fully qualified Dock Blob id, i.e. "did:dock:<SS58 string>"
```
lovesh, line 45: did:dock -> blob:dock

src/modules/schema.js
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  /**
+   * Serializes the object using `getSerializedBlob` and then signs it using the given
+   * polkadot-js pair. The object will be updated with key `signature`. Repeatedly calling it will
+   * keep resetting the `signature` key
+   * @param {any} msg - The message to sign
+   * @param {object} pair - Key pair to sign with
+   */
+  sign(msg, pair) {
```
lovesh, line 85: Why does it take a message to sign? The spec said to use the object state

src/modules/schema.js
```diff
@@ -0,0 +1,189 @@
...
+    this.signature = getSignatureFromKeyringPair(pair, msg);
+    return this;
+  }
+
+  /**
+   * Serializes to JSON for sending to the node, as `Blob`. A full DID is converted to the
+   * hex identifier and signature object is converted to the enum
+   * @returns {object}
+   */
+  toJSON() {
```
lovesh, line 95: The signature is not part of the struct being sent but sent separately

src/modules/schema.js
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  /**
+   * Serializes the schema to a blob object to send to the node
+   * @returns {object}
+   */
+  toBlob(author) {
+    return {
+      id: this.id,
+      blob: stringToHex(JSON.stringify(this.toJSON())),
```
lovesh, line 116: `this.toJSON()` should be `this.schema`. 
Have you checked that the JSON representation is canonical so same fields in different order give same serialization?

src/modules/schema.js
```diff
@@ -0,0 +1,189 @@
...
+   * @returns {Promise<object>}
+   */
+  static async getSchema(id, dockApi) {
+    const hexId = getHexIdentifierFromBlobID(id);
+    const chainBlob = await dockApi.blob.getBlob(hexId);
+    const blobStr = u8aToString(chainBlob[1]);
+    try {
+      const schema = JSON.parse(blobStr);
+      schema.id = id;
+      schema.author = blobHexIdToQualified(u8aToHex(chainBlob[0]));
```
lovesh, line 150: author is not blob.

src/utils/errors/no-blob-error.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * Error thrown when a Blob lookup was successful, but the Blob in question does not exist.
+ * This is different from a network error.
+ */
+export default class NoBlobError extends Error {
+  constructor(blob) {
```
lovesh, line 6: Is `blob` a blob id? If yes, then this name and this.blob should reflect that

src/utils/vc.js
```diff
@@ -295,3 +312,26 @@ export async function isVerifiedPresentation(presentation, challenge, domain, re
...
+/**
+ * The function uses `jsonschema` package to verify that the `credential`'s subject `credentialSubject` has the JSON
+ * schema `schema`
+ * @param {object} credential - The credential to use
+ * @param {object} schema - The schema to use
+ * @returns {Boolean} - Returns promise to an object or throws error
+ */
+export function validateCredentialSchema(credential, schema) {
+  const requiresID = schema.required && schema.required.indexOf('id') > -1;
+  const subjects = credential.credentialSubject.length ? credential.credentialSubject : [credential.credentialSubject];
```
lovesh, line 325: This will make `subjects` array of empty array if `credential.credentialSubject` is empty which is not what you want

src/verifiable-credential.js
```diff
@@ -143,7 +178,12 @@ class VerifiableCredential {
    */
   addSubject(subject) {
     ensureObjectWithId(subject, 'credentialSubject');
-    this.subject = getUniqueElementsFromArray([...this.subject, subject], JSON.stringify);
+    if (!this.credentialSubject || this.credentialSubject.length === 0) {
+      this.credentialSubject = [subject];
+    }
+
+    const subjects = this.credentialSubject.length ? this.credentialSubject : [this.credentialSubject];
```
lovesh, line 185: This will make `subjects` array of empty array if `credential.credentialSubject` is empty which is not what you want

src/modules/schema.js
SamHellawell: you told me to change this in your last bit of feedback, look at what this method does

src/modules/schema.js
SamHellawell: the method description is a bit confusing in this case...

```
   * Serializes to JSON for sending to the node, as `Blob`. A full DID is converted to the
   * hex identifier and signature object is converted to the enum
```

updated this but unclear what you mean in regards to signature above

src/modules/schema.js
SamHellawell: updated to use schema property and throw an error if doesnt exist. do you mean if i serialize to json and back from string will it be in same order they were added in?

src/verifiable-credential.js
SamHellawell: good spot, thanks, resolved in vc class earlier too

src/modules/schema.js
lovesh: Yes, that's confusing. A pair of object and sig is what I was thinking. But you already added a `toBlob` so `toJSON` is redundant. Calling `toBlob` and getting signature should be enough

src/modules/schema.js
lovesh: By canonical, I meant the serialization does not change if the keys of the object are in a different order. One of the 2 documents had mentioned JCS to be used for serialization.

src/modules/schema.js
lovesh: My bad. I wrote blob:dock but should have written did:dock. The author is a DID.

src/modules/schema.js
SamHellawell: toJSON isnt so redundant unless you feel like toblob is all we need, in which case i will rename toblob to tojson. however, in js, tojson methods often serialize the object directly not mutating it in the way toblob does. thoughts?

src/modules/schema.js
SamHellawell: thought was a bit odd, but in this case doesnt that limit us to supporting only dock dids in this class? is that what we want?

SamHellawell: just thinking in terms of interop with other chains/resolvers

lovesh: Its getting schema from the dock chain so the DID will be of Dock

SamHellawell: makes sense, i wasnt sure if our chain-side supported other did methods aswell by way of just input/output

src/modules/schema.js
lovesh: How is `toblob` modifying the object?

src/modules/schema.js
```diff
@@ -0,0 +1,189 @@
...
+
+  /**
+   * Serializes the schema to a blob object to send to the node
+   * @returns {object}
+   */
+  toBlob(author) {
+    return {
+      id: this.id,
+      blob: stringToHex(JSON.stringify(this.toJSON())),
+      author: getHexIdentifierFromDID(author),
```
lovesh, line 119: Any reason you want `this.author` specifically to be overwritable? If not then you can use `this.author`

src/modules/schema.js
SamHellawell: "temporary" implementation (you can see things calling sign passing [1,2,3,4] just to confirm it works) because blob serialization method wasnt implemented when this was. remains a todo, should've had a comment

src/modules/schema.js
SamHellawell: i missed the comment about JCS, will look into this

SamHellawell: added package https://www.npmjs.com/package/json-canonicalize?activeTab=readme 

```Serialization of primitive JSON data types using methods compatible with ECMAScript's JSON.stringify()
Lexicographic sorting of JSON Object properties in a recursive process
JSON Array data is also subject to canonicalization, but element order remains untouched```

seems good, any objections?

Comment (lovesh) [reply src/modules/schema.js]: Thats good. Thanks

File: src/modules/schema.js (hunk 23 above)
Comment (lovesh) [reply src/modules/schema.js:85]: Ok

File: src/modules/schema.js (hunk 30 above)
Comment (SamHellawell) [reply src/modules/schema.js:120]: overlooked this during implementation of toblob method, good spot will adjust

File: src/modules/schema.js (hunk 23 above)
Comment (SamHellawell) [reply src/modules/schema.js]: resolved, however sign method needed to take additional parameter for blob module

File: src/modules/schema.js (hunk 24 above)
Comment (SamHellawell) [reply src/modules/schema.js:95]: perhaps wasn't clear, i dont mean modifying the `Schema` object. what i mean is that `toJSON` should result in a new object that represents the `Schema` object, but as JSON, so that it can be deserialized with a fromJSON method (this is just how most things in javascript go generally). what is returned in `toBlob` is not a representation of the actual schema object

Comment (lovesh) [reply src/modules/schema.js:95]: I see now. So you want to have a to/from JSON and then a toBlob for sending to node. Sounds good

File: src/verifiable-credential.js (hunk 29 above)
Comment (lovesh) [reply src/verifiable-credential.js:185]: This hasn't changed

File: tests/integration/blob.test.js [hunk 31]
```diff
@@ -0,0 +1,161 @@
...
+    expect(!!result).toBe(true);
+
+    const chainBlob = await dock.blob.getBlob(blobId);
+    expect(!!chainBlob).toBe(true);
+    expect(Array.from(chainBlob[1])).toEqual(blobVect);
+  }, 30000);
+
+
+  test('Fails to write blob with size greater than allowed.', async () => {
+    const blobHex = randomAsHex(1025); // Max size is 1024
```
lovesh, line 104: Nitpick: Rather than using 1025, use the constant like MaxBytes+1.

src/verifiable-credential.js
SamHellawell: the fix is above lines 181-183, `credential.credentialSubject` will never be empty

tests/integration/blob.test.js
```diff
@@ -0,0 +1,161 @@
...
+    account = dock.keyring.addFromUri(TestAccountURI);
+    dock.setAccount(account);
+    pair = dock.keyring.addFromUri(firstKeySeed);
+    publicKey = getPublicKeyFromKeyringPair(pair);
+    dockDID = createNewDockDID();
+    keyDetail = createKeyDetail(publicKey, dockDID);
+    txDid = dock.did.new(dockDID, keyDetail);
+    resultDid = await dock.sendTransaction(txDid);
+    didDoc = await dock.did.getDocument(dockDID);
+    blobId = randomAsHex(DockBlobIdByteSize);
```
lovesh, line 61: You can create the DID once in beforeAll, this will make tests run faster as they will do 1 chain write less

src/verifiable-credential.js
lovesh: I see. An easy alternative is `this.credentialSubject = getUniqueElementsFromArray([...this.credentialSubject, subject], JSON.stringify);` without above lines but its fine. Don't want to keep it blocked

tests/integration/schema.test.js
```diff
@@ -0,0 +1,141 @@
...
+    keyDetail = createKeyDetail(publicKey, dockDID);
+    await dock.sendTransaction(dock.did.new(dockDID, keyDetail));
+    blobId = randomAsHex(DockBlobIdByteSize);
+
+    vcInvalid.credentialSchema = {
+      id: blobId,
+      type: 'JsonSchemaValidator2018',
+    };
+
+    // Write invalid format blob
```
lovesh, line 59: A better comment like `blob with invalid JSON-schema format` would be helpful

tests/integration/schema.test.js
```diff
@@ -0,0 +1,141 @@
...
+  }, 120000);
+
+  test('getSchema throws error when no blob exists at the given id.', async () => {
+    await expect(Schema.getSchema(createNewDockBlobId(), dock)).rejects.toThrow(/does not exist/);
+  }, 120000);
+
+  test('Utility method verifyCredential should check if schema is incompatible with the credentialSubject.', async () => {
+    await expect(
+      verifyCredential(
+        vcInvalid, null, false, false, undefined, { notDock: dock },
```
lovesh, line 122: The 3rd arg to `verifyCredential` should be kept true as we are compacting JSON-LD

tests/unit/schema.test.js
```diff
@@ -0,0 +1,204 @@
...
+        schema: expect.anything(),
+      }),
+    );
+  });
+
+  test('toBlob will generate a JSON that can be sent to written with blob module', () => {
+    schema.setAuthor(createNewDockDID());
+    const result = schema.toBlob();
+    console.log('toBlob:', result);
+    expect(result).toMatchObject(
```
lovesh, line 97: It woulld be good to show that this blob can be written to chain.

SamHellawell: this is just a unit test, we have other tests for that in integration

PR #112, opened 2020-05-20
Add schema validation for VerifiablePresentations.

faustow:
Solves https://dock-team.atlassian.net/browse/DCK-22

### DIFF
--- a/src/verifiable-credential.js
+++ b/src/verifiable-credential.js
@@ -178,12 +178,7 @@ class VerifiableCredential {
    */
   addSubject(subject) {
     ensureObjectWithId(subject, 'credentialSubject');
-    if (!this.credentialSubject || this.credentialSubject.length === 0) {
-      this.credentialSubject = [subject];
-    }
-
-    const subjects = this.credentialSubject.length ? this.credentialSubject : [this.credentialSubject];
-    this.credentialSubject = getUniqueElementsFromArray([...subjects, subject], JSON.stringify);
+    this.credentialSubject = getUniqueElementsFromArray([...this.credentialSubject, subject], JSON.stringify);
     return this;
   }
 
--- a/src/verifiable-credential.js
+++ b/src/verifiable-credential.js
@@ -178,7 +178,12 @@ class VerifiableCredential {
    */
   addSubject(subject) {
     ensureObjectWithId(subject, 'credentialSubject');
-    this.credentialSubject = getUniqueElementsFromArray([...this.credentialSubject, subject], JSON.stringify);
+    if (!this.credentialSubject || this.credentialSubject.length === 0) {
+      this.credentialSubject = [subject];
+    }
+
+    const subjects = this.credentialSubject.length ? this.credentialSubject : [this.credentialSubject];
+    this.credentialSubject = getUniqueElementsFromArray([...subjects, subject], JSON.stringify);
     return this;
   }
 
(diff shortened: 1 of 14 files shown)


src/verifiable-credential.js
```diff
@@ -178,12 +178,7 @@ class VerifiableCredential {
    */
   addSubject(subject) {
     ensureObjectWithId(subject, 'credentialSubject');
-    if (!this.credentialSubject || this.credentialSubject.length === 0) {
-      this.credentialSubject = [subject];
-    }
-
-    const subjects = this.credentialSubject.length ? this.credentialSubject : [this.credentialSubject];
-    this.credentialSubject = getUniqueElementsFromArray([...subjects, subject], JSON.stringify);
+    this.credentialSubject = getUniqueElementsFromArray([...this.credentialSubject, subject], JSON.stringify);
```
SamHellawell, line 181: was this a bad merge? the code removed fixes bugs with subject undefined/as object (normally happens with fromjson method creation)

tests/integration/schema.test.js
```diff
@@ -1,119 +1,259 @@
...
-      type: 'JsonSchemaValidator2018',
-    };
-
     // Write invalid format blob
     invalidFormatBlobId = randomAsHex(DockBlobIdByteSize);
-    await dock.sendTransaction(dock.blob.new({
+    await dockApi.sendTransaction(dockApi.blob.new({
       id: invalidFormatBlobId,
-      blob: stringToHex('hello world'),
+      blob: u8aToHex(u8aToU8a('hello world')),
```
SamHellawell, line 54: looks like a bad merge here, should use `stringToHex`

tests/integration/schema.test.js
```diff
@@ -1,119 +1,259 @@
...
       author: getHexIdentifierFromDID(dockDID),
     }, pair), false);
 
     // Write schema blob
     const blobStr = JSON.stringify(exampleSchema);
-    await dock.sendTransaction(dock.blob.new({
+    await dockApi.sendTransaction(dockApi.blob.new({
       id: blobId,
-      blob: stringToHex(blobStr),
+      blob: u8aToU8a(blobStr),
```
SamHellawell, line 62: again should use stringToHex here

src/verifiable-credential.js
faustow: Actually if I remember correctly we ran a test and the deleted line 185 was causing a bug where adding a subject resulted in a two-array subject, one of which was an empty array.

SamHellawell: line 185 does cause that bug, but the code i added above that that is also deleted here fixed it, this has reverted to before it was fixed

SamHellawell: similar fix was implemented for this feedback https://github.com/docknetwork/sdk/pull/110#discussion_r428340774

tests/integration/schema.test.js
faustow:
```suggestion
      blob: stringToHex('hello world'),
```

tests/integration/schema.test.js
```diff
@@ -3,87 +3,128 @@ import { randomAsHex } from '@polkadot/util-crypto';
...
       id: blobId,
       blob: stringToHex(blobStr),
       author: getHexIdentifierFromDID(dockDID),
     }, pair), false);
 
+    // Properly format a keyDoc to use for signing
+    keyDoc = getKeyDoc(
+      dockDID,
+      dockApi.keyring.addFromUri(firstKeySeed, null, 'sr25519'),
+      'Sr25519VerificationKey2020',
```
lovesh, line 73: There is constant declared for this. Better to use that

PR #113, opened 2020-05-22
Tutorial on blobs and schemas.

faustow:
Solves https://dock-team.atlassian.net/browse/DCK-30


tutorials/src/schemas.md
```diff
@@ -0,0 +1,230 @@
...
+<-  true
+```
+
+#### Signing a schema
+Signing a schema can be achieved by calling the `sign` method. It accepts a `msg` param to be signed and a `pair`
+(a keyPair to sign with):
+```javascript
+>   const keyring = new Keyring();
+>   const keypair = keyring.addFromUri(randomAsHex(32), null, 'sr25519');
+>   myNewSchema.sign(myNewSchema.schema, keypair)
```
lovesh, line 143: `sign` needs a keypair and blobModule, `sign(pair, blobModule)`

PR #116, opened 2020-05-22
Schema interface cleanup - rename getblob/getschema to just get, create helper method to write schema to chain to make it easier for users [Jira: DCK-65] 

SamHellawell:


src/modules/schema.js
```diff
@@ -118,6 +118,16 @@ export default class Schema {
     };
   }
 
+  /**
+   * Prepares a transaction to write this schema object to the dock chain using the blob module
+   * @param {object} dock - The dock API
+   * @param {object} pair - The keypair to sign with
+   * @return {object} The extrinsic to sign and send.
+   */
+  writeToChain(dock, pair) {
```
faustow, line 127: This method is returning an extrinsic, can we rename it to something closer to that instead of `writeToChain` since it isn't really writing?

SamHellawell: think best to address with updating method names like `new` etc once we decide 100% if we want these methods to send a tx or not

PR #120, opened 2020-05-26
Update tutorials to reflect new arguments in verify methods

SamHellawell:


tutorials/src/tutorial_ipv.md
```diff
@@ -344,7 +344,7 @@ Here's an example of issuing a Verifiable Credential using DIDs, provided that y
 ```javascript
 const issuerKey = getKeyDoc(issuerDID, dock.keyring.addFromUri(issuerSeed, null, 'ed25519'), 'Ed25519VerificationKey2018');
 await vc.sign(issuerKey);
-const verificationResult = await signedCredential.verify(resolver, true, true, { dock });
+const verificationResult = await signedCredential.verify({ resolver, compactProof: true, forceRevocationCheck: true, revocationApi: { dock } });
```
faustow, line 347: This is good, however we probably need to mention these changes in the other places of this doc where the `verify` methods are introduced. There's https://github.com/docknetwork/sdk/blob/2f84b008a58d50ecc21d9570ae39c939f9525993/tutorials/src/tutorial_ipv.md#verifying-a-verifiable-credential for example where different params are mentioned separately.

SamHellawell: Not sure how i'd address that, do you have any ideas? For me its still a pretty good explanation of properties to pass in the object

faustow: I think just a mention about how all those params go inside an object may be enough. 
Also (and not actually related to this ticket but since we're at it) `schemaApi` param is missing from these intros to the different `verify` methods.

PR #127, opened 2020-05-29
Fix schema example

SamHellawell:


lovesh:
@SamHellawell This should be a hotfix. And lets integrate example runs in the CI as well, i.e. check for non-zero value.

SamHellawell:
Agree on point of CI, had same idea too. Will address in separate PR/branch, should be easy

Issue #125, opened 2020-05-26
Test 'Can create a registry with multiple controllers' is skipped.


bddap:

bddap:
This feature is supported by the node. https://github.com/docknetwork/dock-substrate/blob/42ecfdf2d0efd358857aa720a09194bc8b90379d/runtime/src/revoke.rs#L902

SamHellawell:
Test now passes, perhaps was a previous implementation bug or polkadot issue when first implementing.

SamHellawell:
Test is flaky, fails 50% of the time with badproof error. Still needs more investigation

PR #129, opened 2020-06-01
Unskip and fix "Can create a registry with multiple controllers" test [Jira: DCK-83]

SamHellawell:
Fixes https://github.com/docknetwork/sdk/issues/125

### DIFF
--- a/tests/integration/revocation.test.js
+++ b/tests/integration/revocation.test.js
@@ -64,89 +64,89 @@ describe('Revocation Module', () => {
     await dock.disconnect();
   }, 10000);
 
-  test('Can create a registry with a OneOf policy', async () => {
-    const transaction = dock.revocation.newRegistry(registryId, policy, false);
-    await expect(dock.sendTransaction(transaction)).resolves.toBeDefined();
-    const reg = await dock.revocation.getRevocationRegistry(registryId);
-    expect(!!reg).toBe(true);
-  }, 40000);
+  // test('Can create a registry with a OneOf policy', async () => {
+  //   const transaction = dock.revocation.newRegistry(registryId, policy, false);
+  //   await expect(dock.sendTransaction(transaction)).resolves.toBeDefined();
+  //   const reg = await dock.revocation.getRevocationRegistry(registryId);
+  //   expect(!!reg).toBe(true);
+  // }, 40000);
+  //
+  // test('Can revoke from a registry', async () => {
+  //   const registryDetail = await dock.revocation.getRegistryDetail(registryId);
+  //   expect(!!registryDetail).toBe(true);
+  //
+  //   const lastModified = registryDetail[1];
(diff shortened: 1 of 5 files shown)


tests/integration/revocation.test.js
```diff
@@ -156,6 +156,18 @@ describe('Revocation Module', () => {
...
+    let hasFirstDID = false;
+    let hasSecondDID = false;
+    controllerSet.forEach(controller => {
+      if (controller.toString() === controllerDID) {
+        hasFirstDID = true;
+      } else if (controller.toString() === controllerDIDTwo) {
+        hasSecondDID = true;
+      }
+    });
+    expect(hasFirstDID && hasSecondDID).toBe(true);
```
bddap, line 171: Is this equivalent to checking `new Set(controllerSet) === controllersNew`?

tests/integration/revocation.test.js
```diff
@@ -64,89 +64,89 @@ describe('Revocation Module', () => {
...
+  //   const registryDetail = await dock.revocation.getRegistryDetail(registryId);
+  //   expect(!!registryDetail).toBe(true);
+  //
+  //   const lastModified = registryDetail[1];
+  //   const transaction = dock.revocation.removeRegistry(registryId, lastModified, didKeys);
+  //   await expect(dock.sendTransaction(transaction)).resolves.toBeDefined();
+  //   await expect(dock.revocation.getRegistryDetail(registryId)).resolves.toBeDefined();
+  // }, 40000);
+
+  test('Can create a registry with multiple controllers', async () => {
```
bddap, line 149: Good to go once this section is uncommented.

tests/integration/revocation.test.js
bddap: Maybe more like 
`new Set(controllerSet.map(e => e.toString())) === controllersNew`

SamHellawell: it is equiv, because you cant do `===` check on a Set/array. perhaps it could be written cleaner though https://stackoverflow.com/questions/31128855/comparing-ecma6-sets-for-equality

tests/integration/revocation.test.js
SamHellawell: yeah this wasn't meant to be committed, fixing

SamHellawell:
Test is still flaky here, cant personally figure out why 50% of the time it fails. @lovesh wonder if you have some idea here? Sometimes fails with BadProof error

lovesh:
> Test is still flaky here, cant personally figure out why 50% of the time it fails. @lovesh wonder if you have some idea here? Sometimes fails with BadProof error

Will check in some time

lovesh:
@SamHellawell Please check now

SamHellawell:
works for me, thanks :)

Issue #102, opened 2020-05-08
[Test] Exercise the optional `newController` argument to the `createKeyUpdate` function in the did module


bddap:
```js
test('Can update a DID', async () => {                          
  // send update controller as None
  // assert controller controller is unchanged
  // update controller in did doc
  // assert controller controller is updated
}, 30000);  
```

PR #130, opened 2020-06-01
Add Can update a DID controller test

SamHellawell:
Fixes https://github.com/docknetwork/sdk/issues/102

### DIFF
--- a/tests/integration/did.test.js
+++ b/tests/integration/did.test.js
@@ -3,7 +3,7 @@ import { randomAsHex } from '@polkadot/util-crypto';
 import { DockAPI } from '../../src/api';
 
 import {
-  createNewDockDID,
+  createNewDockDID, getHexIdentifierFromDID,
   createKeyDetail, createSignedKeyUpdate, createSignedDidRemoval,
 } from '../../src/utils/did';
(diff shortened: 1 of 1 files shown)


tests/integration/did.test.js
```diff
@@ -60,6 +60,32 @@ describe('DID Module', () => {
...
+  test('Can update a DID controller', async () => {
+    const pair = dock.keyring.addFromUri(firstKeySeed);
+    const publicKey = getPublicKeyFromKeyringPair(pair);
+
+    // Get the current did controller
+    const originalDoc = await dock.did.getDocument(dockDID);
+    const currentController = getHexIdentifierFromDID(originalDoc.publicKey[0].controller);
+
+    // Send key update without changing controller
+    const [keyUpdateNone, signatureNone] = await createSignedKeyUpdate(dock.did, dockDID, publicKey, pair);
```
lovesh, line 72: The `None` in the name sounds odd. Can you use `let` in place of `const`?

SamHellawell: good point, was taken from andrew's comments in #102 but doesnt make much sense reading back

SamHellawell: also why use let in this case?

Issue #73, opened 2020-04-14
Serialization and deserialization tests for credentials and presentations


lovesh:
Test that credentials and presentations can be converted from and to JSON and signing and verification should work over deserialized objects

PR #131, opened 2020-06-03
Serialize tests [Jira: DCK-79]

SamHellawell:
Fixes https://github.com/docknetwork/sdk/issues/73

### DIFF
--- a/example/resolver.js
+++ b/example/resolver.js
@@ -78,24 +78,22 @@ async function main() {
 
   console.log('Building DIDs list...');
 
+  const dockDID = await createDockDID();
+
   const didsToTest = [
+    dockDID,
     'did:ethr:0xabcabc03e98e0dc2b855be647c39abe984193675',
     'did:work:2UUHQCd4psvkPLZGnWY33L',
     'did:sov:WRfXPg8dantKVubE3HX8pw',
-    'did:ethr:0x3b0BC51Ab9De1e5B7B6E34E5b960285805C41736',
     'did:nacl:Md8JiMIwsapml_FtQ2ngnGftNP5UmVCAUuhnLyAsPxI',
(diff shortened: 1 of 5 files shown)


example/resolver.js
```diff
@@ -78,24 +78,22 @@ async function main() {
 
   console.log('Building DIDs list...');
 
+  const dockDID = await createDockDID();
```
SamHellawell, line 81: had to make a side change here, example was hanging in the CI process. some dids were duplicates and/or not resolving properly, causing a hang but not an error. something in universal resolver side

src/modules/schema.js
```diff
@@ -27,6 +27,24 @@ export default class Schema {
     this.id = id || createNewDockBlobId();
   }
 
+  static fromJSON(json) {
+    const {
+      id, schema, author, ...rest
```
lovesh, line 32: The only extra property is `signature` here, correct? Would we want a schema object with additional properties? If not then we should only check for the ones we know about and skip others.

src/verifiable-presentation.js
```diff
@@ -39,6 +39,42 @@ class VerifiablePresentation {
...
+        vp.addType(type);
+      }
+    }
+
+    const context = rest['@context'];
+    if (context) {
+      vp.context = rest['@context'];
+      delete rest['@context'];
+    }
+
```
lovesh, line 73: Not having a context should raise an error as it won't be a valid JSON-LD object.

src/modules/schema.js
SamHellawell: yeah and we omit signature in our tojson method. good point here, have updated to remove need for rest

src/verifiable-presentation.js
SamHellawell: good point, resolved

Issue #75, opened 2020-04-15
Issuing API requires @context to be an object with a "@context" field


bddap:
I may be wrong, but I suspect jsonld context objects do not need to have a context field. See https://json-ld.org/spec/latest/json-ld/#example-4-context-for-the-sample-document-in-the-previous-section for example of an "@context" with no "@context" field.

This seems to be valid. https://json-ld.org/playground/#startTab=tab-expanded&json-ld=%7B%22%40context%22%3A%7B%22name%22%3A%22http%3A%2F%2Fschema.org%2FName%22%7D%2C%22name%22%3A%22Fondue%20for%20Fun%20and%20Fantasy%22%7D

PR #132, opened 2020-06-03
Dont enforce requirement of @context key in object when setting VP/VC contexts [Jira: DCK-80]

SamHellawell:
Fixes https://github.com/docknetwork/sdk/issues/75

### DIFF
--- a/src/verifiable-presentation.js
+++ b/src/verifiable-presentation.js
@@ -2,13 +2,17 @@ import {
   signPresentation,
   verifyPresentation,
 } from './utils/vc';
-import VerifiableCredential from './verifiable-credential';
+
 import {
-  ensureObjectWithId, ensureObjectWithKeyOrURI, ensureString, ensureURI,
+  ensureObjectWithId,
+  ensureString,
+  ensureURI,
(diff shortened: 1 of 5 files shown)


src/verifiable-presentation.js
```diff
@@ -80,13 +84,28 @@ class VerifiablePresentation {
     return vp;
   }
 
+  /**
+   * Sets the context to the given value, overrding all others
+   * @param {string|object} context - Context to assign
+   * @returns {VerifiablePresentation}
+   */
+  setContext(context) {
```
lovesh, line 92: I had raised this point in earlier reviews as well, let's not duplicate the functions. In this case `setContext` has same logic as in `VerifiableCredential`. Same with `addContext`

SamHellawell: It does have same logic, but are you proposing to write and export a shared function to handle both cases in these classes? Dont think there's a need for such simple logic, if it was more complex or more lines I'd be more inclined to reflector that way

lovesh: Ok

Issue #59, opened 2020-04-09
[Security] When processing jsonld, must not assume "@context"


bddap:
In some places, like `src/utils/vc.js`. Fields from Jsonld documents are dereferenced in a potentially incorrect way.

Examples:

```js
for (let credential of presentation.verifiableCredential) { ... }
```

```js
const regId = credential.credentialStatus.id.slice(DockRevRegQualifier.length);
```

Until this is resolved, our verifier will not be suitable for production.

bddap:
Question: are the checks that vc-js makes sufficient to ensure '@context' is what we expect?

bddap:
Here is an example of the kind of input we should guard against. https://tinyurl.com/ycsyzdcc

Note in the example that "credentialStatus" is reassigned a new meaning. And that "poorlyNamedProperty" is defined to mean "https://www.w3.org/2018/credentials#credentialStatus".

SamHellawell:
>Question: are the checks that vc-js makes sufficient to ensure '@context' is what we expect?

there is context check here https://github.com/digitalbazaar/vc-js/blob/00b7297aa5b7e0e701e8a6db6fb6e98d7606ec0b/lib/vc.js#L498

credentialStatus checks here https://github.com/digitalbazaar/vc-js/blob/00b7297aa5b7e0e701e8a6db6fb6e98d7606ec0b/lib/vc.js#L555 and https://github.com/digitalbazaar/vc-js/blob/00b7297aa5b7e0e701e8a6db6fb6e98d7606ec0b/lib/vc.js#L251

SamHellawell:
Created this PR to check context and status at certain points, since we dont always invoke vcjs. Maybe theres more places this is needed, any ideas @bddap ? https://github.com/docknetwork/sdk/pull/137

bddap:
Looking into it. In the meantime I found this tidbit: https://github.com/json-ld/json-ld.org/issues/391#issuecomment-120099792
"Duplicate context terms are overridden using a most-recently-defined-wins mechanism."

SamHellawell:
marking resolved as of https://github.com/docknetwork/sdk/pull/137

@bddap if any more concerns feel free to open new ticket

PR #137, opened 2020-06-22
Refactor vc.js methods to use expanded credentials and presentations

SamHellawell:
Helps to address some concerns in https://github.com/docknetwork/sdk/issues/59

### DIFF
--- a/src/utils/vc.js
+++ b/src/utils/vc.js
@@ -117,15 +117,11 @@ export async function getCredentialStatuses(expanded) {
   }
 
   statusValues.forEach((status) => {
-    if (status) {
-      if (!status['@id']) {
-        throw new Error('"credentialStatus" must include an id.');
-      }
-      if (!status['@type']) {
-        throw new Error('"credentialStatus" must include a type.');
-      }
-    } else {
-      throw new Error('"credentialStatus" does not exist.');
+    if (!status['@id']) {
+      throw new Error('"credentialStatus" must include an id.');
+    }
+    if (!status['@type']) {
+      throw new Error('"credentialStatus" must include a type.');
     }
   });
 
--- a/src/utils/vc.js
+++ b/src/utils/vc.js
@@ -113,7 +113,7 @@ export async function expandJSONLD(credential) {
 export async function getCredentialStatuses(expanded) {
   const statusValues = jsonld.getValues(expanded, expandedStatusProperty);
   if (statusValues.length === 0) {
-    throw new Error(`Expected ${expandedStatusProperty} after expanding JSON-LD, but not found.`);
+    return statusValues;
   }
 
   statusValues.forEach((status) => {
--- a/src/utils/vc.js
+++ b/src/utils/vc.js
@@ -1,6 +1,7 @@
 import vcjs from 'vc-js';
 import { blake2AsHex } from '@polkadot/util-crypto';
 import { validate } from 'jsonschema';
+import * as jsonld from 'jsonld';
 
 import documentLoader from './vc/document-loader';
 import { isHexWithGivenByteSize } from './codec';
@@ -106,21 +107,40 @@ export function getDockRevIdFromCredential(credential) {
 }
 
 /**
+ * Helper method to ensure credential is valid according to the context
+ * @param credential
+ */
+export async function ensureCorrectJSONLD(credential) {
+  const expanded = await jsonld.expand(credential);
+  return expanded[0];
+}
+
+/**
  * Checks if a credential has a credentialStatus property and it has the properties we expect
  * If status doesnt exist or is invalid, throws an error.
  * @param credential
  */
-export function verifyCredentialStatus(credential) {
-  if (credential.credentialStatus) {
-    if (!credential.credentialStatus.id) {
+export async function verifyCredentialStatus(credential) {
+  const expanded = await ensureCorrectJSONLD(credential);
+  const expandedStatusProperty = 'https://www.w3.org/2018/credentials#credentialStatus';
+  const statusValues = jsonld.getValues(expanded, expandedStatusProperty);
+  if (statusValues.length === 0) {
+    throw new Error('Unable to de-reference "credentialStatus"');
+  }
+
+  const status = statusValues[0];
+  if (status) {
+    if (!status['@id']) {
       throw new Error('"credentialStatus" must include an id.');
     }
-    if (!credential.credentialStatus.type) {
+    if (!status['@type']) {
       throw new Error('"credentialStatus" must include a type.');
     }
   } else {
     throw new Error('"credentialStatus" does not exist.');
   }
+
+  return status;
 }
 
 /**
@@ -144,7 +164,7 @@ export async function checkRevocationStatus(credential, revocationApi) {
   if (!revocationApi.dock) {
     throw new Error('Only Dock revocation support is present as of now.');
   } else {
-    verifyCredentialStatus(credential);
+    await verifyCredentialStatus(credential);
 
     if (!hasDockRevocation(credential)) {
       return { verified: false, error: 'The credential status does not have the format required by Dock' };
@@ -193,6 +213,7 @@ export async function verifyCredential(credential, {
 } = {}) {
   checkCredentialContext(credential);
 
+  await ensureCorrectJSONLD(credential);
   await getAndValidateSchemaIfPresent(credential, schemaApi);
 
   // Run VCJS verifier
@@ -292,6 +313,7 @@ export async function verifyPresentation(presentation, {
 
       // Ensure credential context
       checkCredentialContext(credential);
+      await ensureCorrectJSONLD(credential);
 
       // Check for revocation only if the presentation is verified and revocation check is needed.
       if (isRevocationCheckNeeded(credential.credentialStatus, forceRevocationCheck, revocationApi)) {
--- a/src/utils/vc.js
+++ b/src/utils/vc.js
@@ -145,8 +145,8 @@ function hasDockRevocation(status) {
   const id = status[credentialIDField];
   if (status
     && jsonld.getValues(status, credentialTypeField).includes(RevRegType)
-    && status[credentialIDField].startsWith(DockRevRegQualifier)
-    && isHexWithGivenByteSize(status[credentialIDField].slice(DockRevRegQualifier.length), RevRegIdByteSize)) {
+    && id.startsWith(DockRevRegQualifier)
+    && isHexWithGivenByteSize(id.slice(DockRevRegQualifier.length), RevRegIdByteSize)) {
     return true;
   }
 
--- a/src/utils/vc.js
+++ b/src/utils/vc.js
@@ -75,7 +75,7 @@ export function getSuiteFromKeyDoc(keyDoc) {
 /**
  * Checks if the revocation check is needed. Will return true if `forceRevocationCheck` is true else will check the
  * truthyness of revocationApi. Will return true even if revocationApi is an empty object.
- * @param {object} credential - The credential, expecing a `credentialStatus` field
+ * @param {object} credential - The expanded credential to be checked.
  * @param {boolean} forceRevocationCheck - Whether to force the revocation check.
  * Warning, setting forceRevocationCheck to false can allow false positives when verifying revocable credentials.
  * @param {object} revocationApi - See above verification methods for details on this parameter
--- a/src/utils/vc.js
+++ b/src/utils/vc.js
@@ -25,6 +25,9 @@ export const expandedStatusProperty = `${DEFAULT_CONTEXT_URL}#credentialStatus`;
 export const expandedCredentialProperty = `${DEFAULT_CONTEXT_URL}#verifiableCredential`;
 export const expandedSubjectProperty = `${DEFAULT_CONTEXT_URL}#credentialSubject`;
 export const expandedSchemaProperty = `${DEFAULT_CONTEXT_URL}#credentialSchema`;
+export const credentialIDField = '@id';
+export const credentialContextField = '@context';
+export const credentialTypeField = '@type';
 
 /**
  * @typedef {object} VerifiablePresentation Representation of a Verifiable Presentation.
@@ -93,7 +96,7 @@ export function isRevocationCheckNeeded(credential, forceRevocationCheck, revoca
  */
 export function getDockRevIdFromCredential(credential) {
   // The hash outputs the same number of bytes as required by Dock
-  return blake2AsHex(credential['@id'], RevEntryByteSize * 8);
+  return blake2AsHex(credential[credentialIDField], RevEntryByteSize * 8);
 }
 
 /**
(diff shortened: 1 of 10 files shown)


src/utils/vc.js
```diff
@@ -103,6 +105,34 @@ export function getDockRevIdFromCredential(credential) {
   return blake2AsHex(credential.id, RevEntryByteSize * 8);
 }
 
+/**
+ * Checks if a credential has a credentialStatus property and it has the properties we expect
+ * If status doesnt exist or is invalid, throws an error.
+ * @param credential
+ */
+export function verifyCredentialStatus(credential) {
+  if (credential.credentialStatus) {
```
lovesh, line 114: Isn't the issue raising concern about assuming credential has property `credentialStatus`? 
`status = jsonld.getValues(credential, 'credentialStatus')` is sufficient?

bddap: Maybe not. This is the definition of `getValues`.
https://github.com/digitalbazaar/jsonld.js/blob/7cf2f7155f7d9af6b1c60404cbeeb87788694883/lib/util.js#L304
Not as powerful as we hoped.

lovesh: Maybe of expand (jsonld.expand) first and use the appropriate URI status

bddap:
My understanding is that these changes add the same checks that occur in the latest version of vc-js.
These checks don't actually guard against reassigned properties e.g. https://github.com/docknetwork/sdk/issues/59#issuecomment-639643079

SamHellawell:
I think the checks guard more against mutating context/context not being what is expected, which is what is required to mutate the properties, right? Either way JSON-LD getvalues method looks lacking, but maybe it is wanting an expanded subject

SamHellawell:
pushed some updates, now we will expand the json-ld document to verify credentialStatus is what we expect

src/utils/vc.js
```diff
@@ -65,31 +68,17 @@ export function getSuiteFromKeyDoc(keyDoc) {
...
+ * @param {object} credential - The credential, expecing a `credentialStatus` field
  * @param {boolean} forceRevocationCheck - Whether to force the revocation check.
  * Warning, setting forceRevocationCheck to false can allow false positives when verifying revocable credentials.
  * @param {object} revocationApi - See above verification methods for details on this parameter
  * @returns {boolean} - Whether to check for revocation or not.
  */
-export function isRevocationCheckNeeded(credStatus, forceRevocationCheck, revocationApi) {
-  return !!credStatus && (forceRevocationCheck || !!revocationApi);
+export function isRevocationCheckNeeded(credential, forceRevocationCheck, revocationApi) {
+  return !!credential.credentialStatus && (forceRevocationCheck || !!revocationApi);
```
SamHellawell, line 81: direct de-reference here `credential.credentialStatus` is allowed because we are just checking it exists

bddap: I'm not sure that's correct. If "credentialStatus" is redefined to mean `"http://example.com/otherMeaning"` then this will check whether `"http://example.com/otherMeaning"` property exists.

SamHellawell: it shouldn't matter, because if input is valid (it will be checked after this method is called) then it will verify. if not, it will fail when we expand the document. this just allows us to check if we should expand and process the document properly, without expanding the document

bddap: Hypothetical:

```json5
{
  "@context": [
    "https://www.w3.org/2018/credentials/v1",
    {
      "foo": "https://www.w3.org/2018/credentials#credentialStatus"
    }
  ],
  "@type": "VerifiableCredential",
  "credentialSubject": {
    // ...
  },
  "foo": {
    // some status which, when looked up, indicates "revoked"
  }
}
```

Does/should this credential verify?

SamHellawell: i see your point now, in this case i think the solution requires quite some refactoring in the vc.js methods. Most methods are expecting a non-expanded credential object, and initially I thought just guarding against input we dont expect would work but I see that this should be valid but wont be. Have to also update `verifyPresentation` method to expand the presentation and use expanded credential data.

tests/unit/cred-revocation.test.js
```diff
@@ -1,31 +1,34 @@
-import { isRevocationCheckNeeded } from '../../src/utils/vc';
+import {
+  isRevocationCheckNeeded,
+  expandedStatusProperty
+} from '../../src/utils/vc';
 
 describe('Check isRevocationCheckNeeded', () => {
   test('isRevocationCheckNeeded returns true when credentialStatus is present and forceRevocationCheck is true and revocation API is not given', () => {
-    expect(isRevocationCheckNeeded({}, true, undefined)).toBe(true);
+    expect(isRevocationCheckNeeded({ [`${expandedStatusProperty}`]: {} }, true, undefined)).toBe(true);
```
bddap, line 8: Curious, why the string interpolation?

```
expandedStatusProperty === `${expandedStatusProperty}`
```
right?

tests/unit/cred-revocation.test.js
```diff
@@ -1,31 +1,34 @@
...
   });
 
   test('isRevocationCheckNeeded returns true when credentialStatus is present and forceRevocationCheck is true and revocation API is given', () => {
-    expect(isRevocationCheckNeeded({}, true, { dock: 'Anything' })).toBe(true);
```
bddap, line 9: Hmm. Seems like some of these tests were incorrect before as they weren't passing objects with a credentialStatus property when they claimed to do so.

Good that you corrected them.

src/utils/vc.js
```diff
@@ -327,15 +415,16 @@ export function validateCredentialSchema(credential, schema) {
  * as we support more APIs there are more details associated with each API. Only Dock is supported as of now.
```
bddap, line 415: This function assumes `credential` is in expanded form. Please mention this somwhere in the doc comment.

tests/unit/cred-revocation.test.js
SamHellawell: was using it for more than that at first, have updated

src/utils/vc.js
```diff
@@ -114,17 +174,25 @@ export async function checkRevocationStatus(credential, revocationApi) {
   if (!revocationApi.dock) {
     throw new Error('Only Dock revocation support is present as of now.');
   } else {
-    if (!hasDockRevocation(credential)) {
+    const statuses = await getCredentialStatuses(credential);
+
+    if (!hasDockRevocation(statuses)) {
```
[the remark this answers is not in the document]
SamHellawell: the old method did, but now this supports multiple statuses. see the function body. getCredentialStatuses returns an array

src/utils/vc.js
```diff
@@ -100,7 +93,74 @@ export function isRevocationCheckNeeded(credStatus, forceRevocationCheck, revoca
...
+  statusValues.forEach((status) => {
+    if (status) {
+      if (!status['@id']) {
+        throw new Error('"credentialStatus" must include an id.');
+      }
+      if (!status['@type']) {
+        throw new Error('"credentialStatus" must include a type.');
+      }
+    } else {
+      throw new Error('"credentialStatus" does not exist.');
```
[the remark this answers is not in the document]
SamHellawell: im not sure, can't recall why maybe i included this just to be safe. i think its safe to remove on thinking

src/utils/vc.js
```diff
@@ -100,7 +93,74 @@ export function isRevocationCheckNeeded(credStatus, forceRevocationCheck, revoca
...
+
+/**
+ * Checks if a credential has a credentialStatus property and it has the properties we expect
+ * If status doesnt exist or is invalid, throws an error.
+ * @param expanded
+ */
+export async function getCredentialStatuses(expanded) {
+  const statusValues = jsonld.getValues(expanded, expandedStatusProperty);
+  if (statusValues.length === 0) {
+    throw new Error(`Expected ${expandedStatusProperty} after expanding JSON-LD, but not found.`);
```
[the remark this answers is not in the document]
SamHellawell: good point, updated

src/utils/vc.js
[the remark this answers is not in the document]
bddap: `hasDockRevocation` is the only place that checks:

```js
jsonld.getValues(status, '@type').includes(RevRegType) && id.startsWith(DockRevRegQualifier)
```

We should be checking that every for every status. It's not enough that just one status satisfies that check because we end up doing a revocation lookup for every status in the list.

Consider this credential (look at credentialStatus):

```js
{
  '@context': 'https://www.w3.org/2018/credentials/v1',
  id: 'uuid:0x9b561796d3450eb2673fed26dd9c07192390177ad93e0835bc7a5fbb705d52bc',
  type: [
    'VerifiableCredential',
    'AlumniCredential',
  ],
  issuanceDate: '2020-03-18T19:23:24Z',
  credentialSubject: {
    id: 'did:dock:5GL3xbkr3vfs4qJ94YUHwpVVsPSSAyvJcafHz1wNb5zrSPGi',
    emailAddress: 'johnsmith@example.com',
    alumniOf: 'Example University',
  },
  credentialStatus: [
    {
      id: 'rev-reg:dock:0xf16baed331323c32765faafb7ff15e2f6c07f9cf3a0f96578125e85c71be7574',
      type: 'CredentialStatusList2017',
    },
    {
      id: 'https://revocations.com/registry12345',
      type: 'CredentialStatusList2017',
    },
  ],
  issuer: 'did:dock:5GUBvwnV6UyRWZ7wjsBptSquiSHGr9dXAy8dZYUR9WdjmLUr',
  proof: {
    type: 'Ed25519Signature2018',
    created: '2020-04-22T07:50:13Z',
    jws: 'eyJhbGciOiJFZERTQSIsImI2NCI6ZmFsc2UsImNyaXQiOlsiYjY0Il19..GBqyaiTMhVt4R5P2bMGcLNJPWEUq7WmGHG7Wc6mKBo9k3vSo7v7sRKwqS8-m0og_ANKcb5m-_YdXC2KMnZwLBg',
    proofPurpose: 'assertionMethod',
    verificationMethod: 'did:dock:5GUBvwnV6UyRWZ7wjsBptSquiSHGr9dXAy8dZYUR9WdjmLUr#keys-1',
  },
}
```

src/utils/vc.js
```diff
@@ -170,11 +237,13 @@ export async function verifyCredential(credential, {
   });
 
   // Check for revocation only if the credential is verified and revocation check is needed.
-  if (credVer.verified && isRevocationCheckNeeded(credential.credentialStatus, forceRevocationCheck, revocationApi)) {
-    const revResult = await checkRevocationStatus(credential, revocationApi);
-    // If revocation check fails, return the error else return the result of credential verification to avoid data loss.
-    if (!revResult.verified) {
-      return revResult;
+  if (credVer.verified) {
```
lovesh, line 240: This is functionally equivalent to what was before, why was the nesting introduced in place of the and operator?

src/utils/vc.js
```diff
@@ -327,15 +407,16 @@ export function validateCredentialSchema(credential, schema) {
  * as we support more APIs there are more details associated with each API. Only Dock is supported as of now.
  * @returns {Promise<void>}
  */
-export async function getAndValidateSchemaIfPresent(credential, schemaApi) {
+export async function getAndValidateSchemaIfPresent(credential, schemaApi, context) { // TODO: update to use exanded values
```
lovesh, line 410: Can't this function get context from given credential/presentation

SamHellawell: no because its using expanded credentials, it has no context property

src/utils/vc.js
```diff
@@ -114,17 +166,25 @@ export async function checkRevocationStatus(credential, revocationApi) {
...
+      const regId = status['@id'].slice(DockRevRegQualifier.length);
+
+      // Hash credential id to get revocation id
+      const revocationStatus = await dockAPI.revocation.getIsRevoked(regId, revId); // eslint-disable-line
+      if (revocationStatus) {
+        return { verified: false, error: 'Revocation check failed' };
+      }
     }
+
     return { verified: true };
```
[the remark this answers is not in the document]
SamHellawell: but if the credential has multiple credential statuses that must be checked in future, this is a bad idea. we only support dock revocation for now but do we not intend to support more? what if the dock revocation status is 1st but theres other 2nd, 3rd statuses to look at? we should inform the user that the statuses provided are invalid compared to what we expect, even if there is one valid one

src/utils/vc.js
SamHellawell: i was, before the final updates, doing something else here. dont think nesting is required here anymore

src/utils/vc.js
```diff
@@ -327,15 +415,16 @@ export function validateCredentialSchema(credential, schema) {
  * as we support more APIs there are more details associated with each API. Only Dock is supported as of now.
  * @returns {Promise<void>}
  */
-export async function getAndValidateSchemaIfPresent(credential, schemaApi) {
+export async function getAndValidateSchemaIfPresent(credential, schemaApi, context) { // TODO: update to use exanded values
```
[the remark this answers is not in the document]
SamHellawell: "Do you intend to resolve this TODO before merging" actually it was resolved, just forgot to remove comment :)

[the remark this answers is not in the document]
SamHellawell: most todos i put in code in PRs are to be addressed before merge unless stated, just helps me keep track

src/utils/vc.js
```diff
@@ -100,7 +93,66 @@ export function isRevocationCheckNeeded(credStatus, forceRevocationCheck, revoca
...
+ * @param credential
+ */
+export async function expandJSONLD(credential) {
+  const expanded = await jsonld.expand(credential);
+  return expanded[0];
+}
+
+/**
+ * Checks if a credential has a credentialStatus property and it has the properties we expect
+ * If status doesnt exist or is invalid, throws an error.
```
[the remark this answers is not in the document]
SamHellawell: this was left over from prev functionality, updated docstring

src/utils/vc.js
```diff
@@ -100,7 +93,66 @@ export function isRevocationCheckNeeded(credStatus, forceRevocationCheck, revoca
...
+ * @param expanded
+ */
+export async function getCredentialStatuses(expanded) {
+  const statusValues = jsonld.getValues(expanded, expandedStatusProperty);
+  if (statusValues.length === 0) {
```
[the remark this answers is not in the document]
SamHellawell: good point, it did do something else before refactoring. will remove

src/utils/vc.js
```diff
@@ -100,7 +93,66 @@ export function isRevocationCheckNeeded(credStatus, forceRevocationCheck, revoca
...
+  }
+}
+
+/**
+ * Check if a credential status has a Dock specific revocation
+ * @param status
+ * @returns {Boolean}
+ */
+function hasDockRevocation(status) {
+  const id = status['@id'];
```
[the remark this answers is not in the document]
SamHellawell: i disagree, id argue its cleaner to do

```
if (status
    && jsonld.getValues(status, credentialTypeField).includes(RevRegType)
    && id.startsWith(DockRevRegQualifier)
    && isHexWithGivenByteSize(id.slice(DockRevRegQualifier.length), RevRegIdByteSize)) {
    return true;
  }
```

instead of 

```
  if (status
    && jsonld.getValues(status, credentialTypeField).includes(RevRegType)
    && status[credentialIDField].startsWith(DockRevRegQualifier)
    && isHexWithGivenByteSize(status[credentialIDField].slice(DockRevRegQualifier.length), RevRegIdByteSize)) {
    return true;
  }
```

src/utils/vc.js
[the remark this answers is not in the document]
lovesh: We expect each revocation to be Dock with this `hasDockRevocation` so if there is any other kind, it would fail anyway.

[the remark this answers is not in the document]
SamHellawell: yes but if you moved verified: true into inside the loop, it won't fail if the credential has another status object after the dock one. i think it should, as it would if it had only that status and not a dock one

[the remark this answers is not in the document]
SamHellawell: imo this would result in a false positive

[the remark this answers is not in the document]
SamHellawell: hasDockRevocation now checks individual statuses, rather than all of them before the loop. if this wasnt the case then id agree with you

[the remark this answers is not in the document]
bddap: @lovesh, I don't think there is anything stopping someone from crafting a credential with multiple Dock "credentialStatus"es.

PR #138, opened 2020-06-29
Removal of need to use sendTransaction in SDK, rename sendTransaction to signAndSend

SamHellawell:


src/api.js
```diff
@@ -114,39 +114,56 @@ class DockAPI {
...
+    if (this.customSignTx) {
+      return this.customSignTx(extrinsic, unsubscribe, resolve, reject);
+    }
+
+    const account = this.getAccount();
+    let unsubFunc = null;
+    return extrinsic
+      .signAndSend(account, ({ events = [], status }) => {
+        if (status.isFinalized) {
+          if (unsubscribe && unsubFunc) {
```
lovesh, line 133: If the txn is finalized, `unsubFunc` should never be falsy

src/modules/blob.js
```diff
@@ -79,7 +80,7 @@ class BlobModule {
       // eslint-disable-next-line no-param-reassign
       signature = getSignatureFromKeyringPair(keyPair, serializedBlob);
     }
-    return this.module.new(blob, signature.toJSON());
+    return await this.sendTransaction(this.module.new(blob, signature.toJSON()), false);
```
lovesh, line 83: This means that the transaction once created has to sent. We *must* have a way to create a transaction but not send (was mentioned in the corresponding issue) as that's needed to estimate the fees required (using `paymentInfo`) and sending transactions in parallel (using `utility.batch`). I don't mind if you accomplish that by having a `send` argument (in `new`) that's true by default or have auxiliary function like `writeNew`.

src/modules/did.js
```diff
@@ -15,40 +15,41 @@ class DIDModule {
...
    * @param {string} did - The new DID. Can be a full DID or hex identifier
    * @param {object} keyDetail - `KeyDetail` as expected by the Substrate node
-   * @return {object} The extrinsic to sign and send.
+   * @return {Promise<object>} The extrinsic to sign and send.
    */
-  new(did, keyDetail) {
+  async new(did, keyDetail) {
     const hexId = getHexIdentifierFromDID(did);
-    return this.module.new(hexId, keyDetail);
+    return await this.sendTransaction(this.module.new(hexId, keyDetail));
```
lovesh, line 32: Same comment as above, there must be a way to create transaction without sending.

src/modules/revocation.js
```diff
@@ -12,41 +12,42 @@ class RevocationModule {
...
    * Creating a revocation registry
    * @param {string} id - is the unique id of the registry. The function will check whether `id` is already taken or not.
    * @param {Policy} policy - The registry policy
    * @param {Boolean} addOnly - true: credentials can be revoked, but not un-revoked, false: credentials can be revoked and un-revoked
-   * @return {object} The extrinsic to sign and send.
+   * @return {Promise<object>} The extrinsic to sign and send.
    */
-  newRegistry(id, policy, addOnly) {
-    return this.module.newRegistry(id, {
+  async newRegistry(id, policy, addOnly) {
```
lovesh, line 28: Same comment as above applies to chain writes in this module

src/modules/did.js
```diff
@@ -15,40 +15,41 @@ class DIDModule {
...
    * Updates the details of an already registered DID on the Dock chain.
    * @param {object} keyUpdate - `KeyUpdate` as expected by the Substrate node
    * @param {Signature} signature - Signature from existing key
-   * @return {object} The extrinsic to sign and send.
+   * @return {Promise<object>} The extrinsic to sign and send.
    */
-  updateKey(keyUpdate, signature) {
-    return this.module.updateKey(keyUpdate, signature.toJSON());
+  async updateKey(keyUpdate, signature) {
+    return await this.sendTransaction(this.module.updateKey(keyUpdate, signature.toJSON()));
```
lovesh, line 42: Same comment as above applies to chain writes in this module

src/modules/schema.js
```diff
@@ -137,10 +137,10 @@ export default class Schema {
    * Prepares a transaction to write this schema object to the dock chain using the blob module
    * @param {object} dock - The dock API
    * @param {object} pair - The keypair to sign with
-   * @return {object} The extrinsic to sign and send.
+   * @return {Promise<object>} The extrinsic to sign and send.
    */
-  writeToChain(dock, pair) {
-    return dock.blob.new(this.toBlob(), pair);
+  async writeToChain(dock, pair) {
+    return await dock.blob.new(this.toBlob(), pair, false);
```
lovesh, line 143: You can omit the false.

src/api.js
```diff
@@ -114,39 +114,56 @@ class DockAPI {
...
+      });
+  }
+
   /**
    * Helper function to send transaction
    * @param {object} extrinsic - Extrinsic to send
    * @param {Boolean} unsubscribe - Should we unsubscribe from the transaction after its finalized
    * @return {Promise}
    */
   async sendTransaction(extrinsic, unsubscribe = true) {
```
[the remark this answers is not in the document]
SamHellawell: we already pass unsubscribe = false for some methods

src/api.js
```diff
@@ -30,12 +30,12 @@ import {
...
   /**
-   * Skeleton constructor, does nothing yet
+   * Creates a new instance of the DockAPI object, call init to initialize
+   * @param {function} [customSignTx] - Optional custom transaction sign method
    * @constructor
-   * @param {string} address - WebSocket RPC endpoint
    */
-  constructor(address = null) {
-    this.address = address;
+  constructor(customSignTx) {
```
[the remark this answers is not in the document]
SamHellawell: yeah im aware this solution isnt the best, either needs proper documenting or re-thinking. ideally i wouldn't want them to have to send the tx too, just sign it. need to see if polkadot has any methods for sending/signing tx's separately

src/modules/blob.js
SamHellawell: i was thinking that in this case user could hook the custom sign tx's in, it is hard to think of a good solution for that without seeing a use case where the tx is separated from sending

SamHellawell: at this point im leaning towards a new argument in new that is true by default, otherwise return the tx

src/modules/schema.js
SamHellawell: can't recall 100% why for blob module we are passing false here, maybe something fausto encountered as he developed it. i'll see what happens if we remove this

src/modules/blob.js
lovesh: Its fine if you are short on time but keep in mind the return type becomes ugly now.

src/api.js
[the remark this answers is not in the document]
SamHellawell: have removed need for unsubscribe, initially i think it was included just to support it incase, not sure why it was passed as false for some methods. all seems to work assuming true, so think we can safely remove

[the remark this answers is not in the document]
lovesh: Can you tell why, the promise resolves only when txn in finalized after which there is no need of keeping it subscribed. On a different note, can we give an option to either return on finalize or inBlock (we can default to finalize)?

src/modules/schema.js
SamHellawell: seems to be fine when removed

src/modules/blob.js
SamHellawell: yeah on further thought not so happy with that, would rather make a new method to create tx and keep current ones using sendtransaction. see https://github.com/docknetwork/sdk/pull/138/commits/64df79b14625068a21db1ebd791d0d305b8f9683

src/modules/did.js
```diff
@@ -15,42 +15,73 @@ class DIDModule {
...
 
   /**
-   * Creates a new DID on the Dock chain.
+   * Creates transaction to create a new DID on the Dock chain.
    * @param {string} did - The new DID. Can be a full DID or hex identifier
    * @param {object} keyDetail - `KeyDetail` as expected by the Substrate node
    * @return {object} The extrinsic to sign and send.
    */
-  new(did, keyDetail) {
+  createNewTx(did, keyDetail) {
```
lovesh, line 30: I would have preferred not to have `create` as that's common for methods creating transactions but put the important part first like `newTx` or `updateKeyTx` but just expressing my opinion, not requiring a change

PR #142, opened 2020-07-17
Update blob module to auto detect value type so user doesnt need to import helper methods

SamHellawell:
Also should try to auto detect input type in get method, eg: passing json will result in json from get method without need for user to parse it

### DIFF
--- a/src/modules/blob.js
+++ b/src/modules/blob.js
@@ -1,4 +1,6 @@
 import { encodeAddress, randomAsHex } from '@polkadot/util-crypto';
+import { stringToHex } from '@polkadot/util';
+
 import { getSignatureFromKeyringPair, getStateChange } from '../utils/misc';
 import { isHexWithGivenByteSize, getHexIdentifier } from '../utils/codec';
 import NoBlobError from '../utils/errors/no-blob-error';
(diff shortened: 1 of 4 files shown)


src/modules/blob.js
```diff
@@ -72,15 +74,31 @@ class BlobModule {
    * @return {object} The extrinsic to sign and send.
    */
   createNewTx(blob, keyPair = undefined, signature = undefined) {
+    let value = blob.blob;
+    if (!value) {
+      throw new Error('Blob must have a value!');
+    }
+
+    if (typeof value === 'object' && !Array.isArray(value) && !(value instanceof Uint8Array)) {
```
lovesh, line 82: I see what you are doing here but it would still be helpful to document what are you expecting and what conversions are you doing. 
I think we should accept `Uint8Array` or hex string only and keep the caller responsible for ensuring that. A dangerous example is caller passing the blob as an array with integers > 255. Polkadot's serialization will truncate the bigger integers to fit in a byte as the requires a byte array. So better to just check for `Uint8Array` and hex only

SamHellawell: fair enough observation, my initial idea with allowing normal arrays was just to prevent user from having to construct a typed array, but we can support only typed i dont think it'll make a huge difference for usability

SamHellawell:
@lovesh have addressed feedback and fixed tests

src/modules/blob.js
```diff
@@ -72,15 +74,33 @@ class BlobModule {
...
+    let value = blob.blob;
+    if (!value) {
+      throw new Error('Blob must have a value!');
+    }
+
+    if (value instanceof Uint8Array) {
+      value = [...value];
+    } else if (typeof value === 'object') {
+      value = stringToHex(JSON.stringify(value));
+    } else if (typeof value === 'string' && value.substr(0, 2) !== '0x') {
```
lovesh, line 86: You might end up with invalid hex starting with 0x, like odd length or some non-hex chars, better to use the available utility `isHexWithGivenByteSize` in place of `(typeof value === 'string' && value.substr(0, 2) !== '0x')` and move that check upwards

SamHellawell: issue with `isHexWithGivenByteSize` is that it requires a byte size, cant blobs be between a min and max length?

SamHellawell: nvm, i just read docs on `isHexWithGivenByteSize`, notice that byteSize can be undefined - not so clear by naming of function

SamHellawell: updated, no need to move check above as im trying to exclude whats already hex here

PR #152, opened 2020-08-13
exercise Master functionality from client

bddap:
This pr introduces integration tests for the Master module on the node. The tests are intended to be run against node at the head of `poa-1` in the dock-sdk repo: e4c5e318cd78d58a4af8de8907e4d90705138b66.

As mentioned in the comment below, this branch should be rebased onto `poa-1` of this repo. TODO

This branch has failing tests and lints but not related to the changes introduced. I think those failing tests and lints should be fixed separately, after merging.

You'll notice `master.test.js` does not use functions exported by the sdk.

### DIFF
--- a/src/types.json
+++ b/src/types.json
@@ -90,6 +90,15 @@
     "locked_reward": "Option<Balance>",
     "unlocked_reward": "Option<Balance>"
   },
+  "Payload": {
+    "proposal": "Vec<u8>",
+    "round_no": "u64"
+  },
+  "Membership": {
+    "members": "BTreeSet<Did>",
+    "vote_requirement": "u64"
+  },
+  "PMAuth": "BTreeMap<Did, DidSignature>",
   "StateChange": {
     "_enum": {
       "KeyUpdate": "KeyUpdate",
@@ -97,7 +106,8 @@
       "Revoke": "Revoke",
       "UnRevoke": "UnRevoke",
       "RemoveRegistry": "RemoveRegistry",
-      "Blob": "Blob"
+      "Blob": "Blob",
+      "MasterVote": "Payload"
     }
   }
 }
--- a/tests/integration/master.test.js
+++ b/tests/integration/master.test.js
@@ -1,0 +1,155 @@
+import types from '../../src/types.json';
+import { ApiPromise, WsProvider, Keyring } from '@polkadot/api';
+import { cryptoWaitReady } from '@polkadot/util-crypto';
+import { FullNodeEndpoint, TestAccountURI, TestKeyringOpts } from '../test-constants';
+import { assert, u8aToU8a } from '@polkadot/util';
+import Bytes from '@polkadot/types/primitive/Bytes';
+
+describe('Master Module', () => {
+  // node client
+  let nc;
+
+  beforeAll(async (done) => {
+    nc = await connect();
+    done();
+  }, 40000);
+
+  afterAll(async () => {
+    await nc.disconnect();
+  }, 10000);
+
+  test('Can authenticate and execute a root call', async () => {
+    let key = add_len_prefix([0xaa, 0xbb]);
+    let val = add_len_prefix([0xcc, 0xdd]);
+
+    // construct call
+    let call = nc.tx.system.setStorage([(key, val)]);
+
+    // sign call
+    let payload = {
+      proposal: nc.createType('Call', call).toU8a(),
+      round_no: await nc.query.master.round(),
+    };
+    let encoded_state_change = nc.createType('StateChange', { MasterVote: payload }).toU8a();
+
+    // submit without votes
+    await (async () => {
+      let res = await sign_send_tx(nc.tx.master.execute(call, []));
+      assert(res.status.isFinalized, "transaction was not included in block");
+      let sto = await nc.rpc.state.getStorage(key);
+      assert(sto.isNone, "storage item should not have been set");
+    })();
+
+    // // submit with invalid votes
+    // await (async () => {
+    //   let res = await sign_send_tx(nc.tx.master.execute(call, ));
+    //   assert(res.status.isFinalized, "transaction was not included in block");
+    //   let sto = await nc.rpc.state.getStorage(key);
+    //   assert(sto.isNone, "storage item should not have been set");
+    // })();
+
+    // submit with valid votes
+    await (async () => {
+      const bts = str => new TextEncoder("utf-8").encode(str);
+      const did_to_key = [
+        [
+          bts("Alice\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"),
+          await keypair(bts("Alicesk\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"))
+        ],
+        [
+          bts("Bob\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"),
+          await keypair(bts("Bobsk\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"))
+        ],
+        [
+          bts("Charlie\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"),
+          await keypair(bts("Charliesk\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"))
+        ],
+      ];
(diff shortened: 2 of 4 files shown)


lovesh:
This should go in the branch poa-1 and should have been based off it

bddap:
These tests are now working against the current head of `poa-1` (node repo). All that is left is to rebase onto `poa-1` (this repo), make sure `poa-1` (this repo) is testing against the latest node, and review this PR.

src/types.json
```diff
@@ -75,28 +75,39 @@
...
+    "locked_reward": "Option<Balance>",
+    "unlocked_reward": "Option<Balance>"
   },
+  "Payload": {
+    "proposal": "Vec<u8>",
+    "round_no": "u64"
+  },
+  "Membership": {
+    "members": "BTreeSet<Did>",
+    "vote_requirement": "u64"
```
lovesh, line 99: Nitpick: Does it seem practical that `vote_requirement` will ever be over u8 (>255).

tests/integration/master.test.js
```diff
@@ -0,0 +1,255 @@
...
+    ];
+    await master_set_storage(nc, key, val, did_to_key);
+    let sto = await nc.rpc.state.getStorage(key);
+    let u8a = sto.unwrap();
+    expect(u8a_hex(u8a)).toEqual(val);
+  }, 20000);
+});
+
+// connect to running node
+async function connect() {
```
lovesh, line 134: `DockAPI`'s `init` does this so a call like `(new DockAPI()).init(FullNodeEndpoint)` would be equivalent.

tests/integration/master.test.js
```diff
@@ -0,0 +1,255 @@
...
+// load a DID kp from secret
+async function keypair(seed) {
+  assert(seed instanceof Uint8Array, "wrong type");
+  await cryptoWaitReady();
+  let keyring = new Keyring({ type: 'sr25519' });
+  let key = keyring.addFromSeed(seed);
+  return key
+}
+
+async function sign_send_tx(extrinsic) {
```
lovesh, line 157: This function is a duplicate of `signAndSend` from DockAPI object. The following lines should do what this function does
```
// dock is DockAPI object
dock.setAccount(key);
dock.signAndSend(extrinsic)
```
The naming should follow camelCase.

tests/integration/master.test.js
```diff
@@ -0,0 +1,255 @@
...
+      reject(error);
+    }
+
+    return this;
+  });
+
+  return await promise;
+}
+
+async function get_test_account_key() {
```
lovesh, line 189: camelCase.
`addFromUri` will work for seed as well so `keypair` function isn't needed.

tests/integration/master.test.js
```diff
@@ -0,0 +1,255 @@
...
+  await cryptoWaitReady();
+  let keyring = new Keyring(TestKeyringOpts);
+  let key = keyring.addFromUri(TestAccountURI);
+  return key
+}
+
+// represent a Uint8Array as hex with a "0x" prefix
+function u8a_hex(bs) {
```
lovesh, line 197: Polkadot-js's util library has a function `u8aToHex` to do this conversion.

tests/integration/master.test.js
```diff
@@ -0,0 +1,255 @@
...
+  assert(bs instanceof Uint8Array, "wrong type");
+  return '0x' + [...bs].map(bt => ('0' + bt.toString(16)).slice(-2)).join('');
+}
+expect(u8a_hex(new Uint8Array([]))).toEqual('0x');
+expect(u8a_hex(new Uint8Array([0x01]))).toEqual('0x01');
+expect(u8a_hex(new Uint8Array([0xaa, 0xbb]))).toEqual('0xaabb');
+
+function randomish_u8a() {
```
lovesh, line 205: Whats the purpose this function? Are you looking for a pseudorandom 32 byte UintArray or with something slightly different? For the former, polkadot-js has `randomAsU8a`

tests/integration/master.test.js
```diff
@@ -0,0 +1,255 @@
...
+  key, // hex encoded bytes
+  val, // hex encoded bytes
+  did_to_key, // list of (did, key) pairs with which to vote
+) {
+  let call = nc.tx.system.setStorage([[key, val]]); // this is a root-only extrinsic
+  let payload = {
+    proposal: [...nc.createType('Call', call).toU8a()],
+    round_no: await nc.query.master.round(),
+  };
+  let encoded_state_change = nc.createType('StateChange', { MasterVote: payload }).toU8a();
```
lovesh, line 224: This can be achieved through `getBytesForStateChange` as well.

tests/integration/master.test.js
```diff
@@ -0,0 +1,255 @@
...
+
+  let votes = new Map();
+  for (let [did, key] of dtk_sorted) {
+    votes.set(did, { Sr25519: key.sign(encoded_state_change) });
+  }
+  await sign_send_tx(nc.tx.master.execute(call, votes));
+}
+
+/// convert a string to utf-8 encoded bytes
+function bts(str) {
```
lovesh, line 237: polkadot-js's `stringToU8a` would do the same thing.

tests/integration/master.test.js
```diff
@@ -0,0 +1,255 @@
...
+) {
+  let call = nc.tx.system.setStorage([[key, val]]); // this is a root-only extrinsic
+  let payload = {
+    proposal: [...nc.createType('Call', call).toU8a()],
+    round_no: await nc.query.master.round(),
+  };
+  let encoded_state_change = nc.createType('StateChange', { MasterVote: payload }).toU8a();
+
+  let dtk_sorted = [...did_to_key];
+  dtk_sorted.sort(([dida, _a], [didb, _b]) => compare_u8a_32(dida, didb));
```
lovesh, line 227: If you don't care about performance and want to reduce code, taking did as hex and calling `sort` should also work, avoiding the need for `compare_u8a_32`

src/types.json
bddap: u8 or u16 would have been fine too. It's set to u64 in the runtime.

bddap:
Thanks for the review @lovesh , I will see to your comments soon.

tests/integration/master.test.js
bddap: Thanks. I'm aware. This module intentionally does not depend on sdk (except for types.json).

tests/integration/master.test.js
bddap: Same comment as above.

tests/integration/master.test.js
bddap: noted

bddap:
@lovesh I removed the "scale encoding of call" test in https://github.com/docknetwork/sdk/pull/152/commits/78253d88b1a7d02c56b229087543695e2b6fc747 .
- The test was failing because is was expecting incorrect results.
- It was not related to the problem being tested.
- It wasn't future-proof, changes to the system pallet (adding new calls) would result in the test failing.
- It wasn't a useful test.

tests/integration/master.test.js
lovesh: Whats the concern with using the SDK?

bddap: Decoupling from SDK makes this test portable and self-contained. It might be best to eventually move e2e node tests over to the dock-substrate repo, in that case depending on the sdk would cause problems related to circular dependency.

tests/integration/master.test.js
bddap: Switched to camelCase in 974bafd.

tests/integration/master.test.js
bddap: Thanks. Removed in `u8a_hex` in favor of `u8aToHex` in 974bafd.

tests/integration/master.test.js
bddap: Thanks. Removed in `bts` in favor of `stringToU8a` in 974bafd.

tests/integration/master.test.js
bddap: done: 974bafd

tests/integration/master.test.js
bddap: That's correct. removed `randomish_u8a` in favor of `randomAsU8a` in 974bafd.

tests/integration/master.test.js
bddap: Switched to camelCase in 974bafd.

tests/integration/master.test.js
bddap: Thanks for the suggestion. Merged `getTestAccountKey()` and `keypair()` in 1b7072a

tests/integration/master.test.js
bddap: I'd like to keep this one as-is.

tests/integration/master.test.js
lovesh: Makes sense

lovesh (approved):

PR #154, opened 2020-08-17
Master vote scripts

bddap:
Adds two scripts, one for voting on a proposal and one for executing a proposal with votes. `./scripts/master_voting/vote.js` and `./scripts/master_voting/submit.js` respectively.

Both scripts do some validation to keep users on the right track. For example `vote.js` checks the current voting round and gets snippy with the user if the one they entered the wrong one. `submit.js` won't let the you submit a proposal if the votes are invalid.

`vote.js` has a dry-run mode that is mostly for user comfort. In dry-run mode `vote.js` prints the proposal but doesn't sign it.

`vote.js` makes the voter explicitly enter the voting round in order to force them to know it.

[Node side](https://github.com/docknetwork/dock-substrate/blob/b0c968d9da553cb28083ea33d467d0841199a3be/runtime/src/master.rs) for reference.

Requesting @SamHellawell to review because these scripts would be usable as reference when [implementing a gui version](https://dock-team.atlassian.net/jira/software/projects/DCK/boards/2?selectedIssue=DCK-339).


scripts/master_voting/example_votes.json
```diff
@@ -0,0 +1,10 @@
+[
```
bddap, line 1: These votes are valid over `example_proposal.json` but only when the on-chain voting round is `0`.

scripts/master_voting/vote.js
```diff
@@ -0,0 +1,105 @@
...
+`;
+
+main().catch(e => {
+  console.error(e);
+  process.exit(1);
+}).then(_ => {
+  process.exit(0);
+});
+
+async function main() {
```
bddap, line 33: You'll notice the program structure of this script is slightly different from that of `scripts/master_voting/submit.js`. There isn't a great technical reason for the difference. If it's a problem let me know and I'll rewrite `scripts/master_voting/vote.js` in the style of `scripts/master_voting/submit.js`.

SamHellawell (approved):

SamHellawell:
r.e linter error, it should go away on merge i think. if not i will fix it post-merge - due to a polkadot version mismatch

PR #159, opened 2020-08-21
implement keyupdate script

bddap:
Adds a script to update your own did key. This is needed in order to update the keys of master members on the vulcan testnet.

To save time I didn't add any argument parsing. Users will need to edit this script manually for now.

edit: I added argument parsing.

### DIFF
--- a/scripts/change_did_key.js
+++ b/scripts/change_did_key.js
@@ -1,21 +1,41 @@
+/// This script assumes that all keys are sr25519.
+///
+/// Warning: This script only supports sr25519 keys. Do not use it to set an non-sr25519 public key.
+///          If you do, you will lose control of your DID.
+
 import { Keyring } from '@polkadot/api';
 import { cryptoWaitReady } from '@polkadot/util-crypto';
-import { u8aToHex } from '@polkadot/util';
+import { u8aToHex, assert } from '@polkadot/util';
 import { createSignedKeyUpdate } from '../src/utils/did';
 import { getPublicKeyFromKeyringPair } from '../src/utils/misc';
 import { DockAPI, PublicKeySr25519 } from '../src/api';
 import b58 from 'bs58';
 
-/// This script is not configurable via command line. You will need to maually edit `main()`
-/// to input your own values.
-///
+require('dotenv').config();
+
 /// Use (with node running locally):
 ///
 /// ```bash
 /// cd sdk
 /// yarn
-/// npx babel-node ./scripts/change_did_key.js
+/// npx babel-node ./scripts/change_did_key.js args..
 /// ```
+
+const USAGE = `\
+npx babel-node ./scripts/change_did_key.js <did> <new_pk>
+  where
+    <did> is a 32 byte, 0x prefixed hex string.
+    <new_pk> is a 32 byte, 0x prefixed hex string.
+
+env vars:
+  - FullNodeEndpoint: websocket Url of a trusted node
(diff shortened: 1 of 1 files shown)


scripts/change_did_key.js
```diff
@@ -0,0 +1,113 @@
...
+  };
+  /// Some arbitray sr25519 keypair.
+  const TEST_KP = {
+    sk: '0x6b9abb2153e0e3989bb53869b4a88806941a0a86cefa08a7977aff73379df734',
+    pk: '0xc081a33d27d66ceda4e9f634f2f3a1b3b7f4ad898662036416250862ffae623f',
+  };
+
+  // set key to TEST_KP
+  await updateDIDKey({
+    nodeWsUrl: 'ws://localhost:9944',
```
lovesh, line 44: I have read the note in PR but it should not be too much work reading from .env file like other scripts, eg. https://github.com/docknetwork/sdk/blob/poa-1/scripts/add_session_key_with_sudo.js#L9

scripts/change_did_key.js
```diff
@@ -0,0 +1,113 @@
...
+    did: ALICE_DID,
+    currentSk: ALICE_KP.sk,
+    newPk: TEST_KP.pk,
+    payerSk: "//Bob"
+  });
+
+  // set it back to the default
+  await updateDIDKey({
```
lovesh, line 52: Why is the update reverted?

lovesh (changes requested):

scripts/change_did_key.js
bddap: For testing. It makes it so you can run this script repeatedly without needing to `purge-chain`.

lovesh: But that would require the actual user to comment out this before he runs the script?

scripts/change_did_key.js
bddap: added argument parsing in 29882d2

scripts/change_did_key.js
bddap: This function call is no longer present.

scripts/change_did_key.js
```diff
@@ -0,0 +1,150 @@
...
+
+/// Use (with node running locally):
+///
+/// ```bash
+/// cd sdk
+/// yarn
+/// npx babel-node ./scripts/change_did_key.js args..
+/// ```
+
+const USAGE = `\
```
lovesh, line 24: This is a good change as it does not require editing the script anymore.

scripts/change_did_key.js
```diff
@@ -0,0 +1,150 @@
...
+npx babel-node ./scripts/change_did_key.js <did> <new_pk>
+  where
+    <did> is a 32 byte, 0x prefixed hex string.
```
lovesh, line 27: You don't need to require 32 bytes. The places where this is used, i.e. in `createSignedKeyUpdate` and `getSr25519PkHex` both accept hex or qualified like `did:dock:..`. You can use `getHexIdentifierFromDID` to validate the script arg for that. People creating a DID will usually have a qualified did.

lovesh (approved):

scripts/change_did_key.js
bddap: Used `getHexIdentifierFromDID` to allow fully qualified dids. 754c03c

Instead of propagating the possibly fully qualified DID deeper into the program I parse it using `getHexIdentifierFromDID` and pass on the parsed value. Behavior is the same.

PR #160, opened 2020-08-28
Allow waiting for block inclusion, passing nonce using params and dec…

lovesh:
…ouple API for PoA types

Signed-off-by: lovesh <lovesh.bond@gmail.com>

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -54,19 +54,9 @@ class DockAPI {
 
     this.address = address || this.address;
 
-    // Polkadot-js needs these extra type information to work. Removing them will lead to
-    // an error. These were taken from substrate node frontend template.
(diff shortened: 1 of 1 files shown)


src/api.js
```diff
@@ -54,19 +54,9 @@ class DockAPI {
 
     this.address = address || this.address;
 
-    // Polkadot-js needs these extra type information to work. Removing them will lead to
-    // an error. These were taken from substrate node frontend template.
-    const extraTypes = {
```
lovesh, line 59: Removed them as they are already present in types.json

SamHellawell: good spot

SamHellawell (approved):

PR #169, opened 2020-09-15
Create a composite claim soundness checker 

bddap:
This checker is run by the verifier. It adds the ability to accept proofs of composite claims.

`acceptCompositeClaims` in `./src/util/cd.js` is central to this pr. It returns a claimgraph which includes the the claims which were proven.

 Proof validation is a separate step from validation of a presentation. The two steps are nicely unrelated to each other and can even be done at the same time with `Promise.all()` if the user desires. The separation allows the user control over validation setting, like `forceRevocationCheck`, without the claim deduction interface needing to know about the validation interface.

When encountering an invalid proof my options were to throw an error or to express the error in the return type e.g.
`{ Ok: results } | { Err: invalidReason }`. I currently throw an error because I think that is idiomatic in js. I can change this if desired.

As a UX decision, I intentionally left out an explicit `checkSoundness` function, though it's very simple to implement one. There is `validate` and `acceptCompositeClaims`. If both of those succeed, then the claims returned by `acceptCompositeClaims(_, rules)` are **true** with respect to **rules**.

This pr also adds the closely related ability to for the holder to generate a proof of composite claims for inclusion in a presentation. see `proveCompositeClaims()` in `./src/util/cd.js`.


.github/workflows/test.yml
```diff
@@ -8,7 +8,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
         with:
-          node-version: '10.x'
+          node-version: '12.x'
```
SamHellawell, line 11: should be version 13.x

package.json
```diff
@@ -71,9 +71,11 @@
     "@polkadot/types": "1.34.1",
     "@polkadot/util": "3.4.1",
     "@polkadot/util-crypto": "3.4.1",
+    "assert": "^2.0.0",
```
SamHellawell, line 74: im pretty sure polkadot has some assert lib, maybe worth using that instead of this

tests/network-cache.js
```diff
@@ -0,0 +1,76 @@
+export default {
```
SamHellawell, line 1: would advise creating a subfolder or moving out of initial tests dir since its not a test, filename is a bit misleading if you dont know we use *.test.js

SamHellawell (changes requested):
looks good to me for most part besides CI issues and few comments. i don't think i fully understand the logic enough to review that part, but I understand whats going on here mostly

tests/network-cache.js
bddap: Like `./test/network-cache/network-cache.js`? Or `./test/constants/network-cache.js`

Note we already have other constants declared in that directory and this is following a similar pattern.

```
./sdk> ls tests/
example-credential.js  example-schema.js  integration/  network-cache.js  poa/  test-constants.js  unit/
```

.github/workflows/test.yml
bddap: Changed to `13.x` in d9b4eba

package.json
bddap: Removed the `assert` dependency in favor of polkadot's version. d9b4eba

bddap:
> When encountering an invalid proof my options were to throw an error or to express the error in the return type e.g.
{ Ok: results } | { Err: invalidReason }. I currently throw an error because I think that is idiomatic in js. I can change this if desired.

@SamHellawell does it make sense to do it this way?

SamHellawell:
@bddap yeah that seems fine

tests/network-cache.js
SamHellawell: i think we can leave it for now and clean up later

SamHellawell (approved):

PR #174, opened 2020-09-23
Accept external types and RPC definitions in Api

lovesh:


src/api.js
```diff
@@ -59,18 +61,43 @@ class DockAPI {
 
     this.address = address || this.address;
 
+    // If RPC methods given, use them else set it to empty object.
+    let rpc = chainRpc || {};
+
+    // If using PoA module, extend the RPC methods with PoA specific ones.
+    if (loadPoaModules) {
+      rpc = Object.assign(rpc, {
+        poa: {
```
SamHellawell, line 70: Why are we hard coding this instead of pulling from node like with other rpc methods?

lovesh: Because these are custom for us. I think we can avoid by submitting them in Polkadot-js but this is a short term change so avoiding.

SamHellawell: i still dont quite understand why the node's metadata doesnt include these rpc calls and why we have to supply their info manually here

lovesh: Because RPC's are not part of node's metadata or the runtime. Exposing an RPC or not is a node's choice (it has to modify our code to not expose though). As an example, [here](https://github.com/polkadot-js/api/blob/master/packages/types/src/interfaces/payment/definitions.ts) is the payment RPC defined in polkadot-js that lets you query fees and weight which is by default loaded but it is not learnt from the node.

SamHellawell (approved):

PR #178, opened 2020-09-28
use latest vc-js package not our fork, add jsonld-signatures to resolutions

SamHellawell:


src/utils/type-helpers.js
```diff
@@ -1,4 +1,4 @@
-import vcjs from 'vc-js';
+import vcjs from 'vc-js/lib';
```
SamHellawell, line 1: we import like this so that it doesnt use the precompiled dist vcjs version which includes jsonld signatures by default

package.json
```diff
@@ -49,7 +52,7 @@
     "build": "rollup -c",
     "watch": "rollup -c -w",
     "test": "jest --verbose ./tests/unit",
-    "test-integration": "NODE_ENV=production jest --verbose --runInBand ./tests/integration",
+    "test-integration": "NODE_ENV=production jest --verbose --runInBand ./tests/integration/issuing",
```
lovesh, line 55: Why skip the other tests?

src/utils/vc.js
```diff
@@ -232,6 +232,9 @@ export async function verifyCredential(credential, {
     suite: [new Ed25519Signature2018(), new EcdsaSepc256k1Signature2019(), new Sr25519Signature2020()],
     documentLoader: documentLoader(resolver),
     compactProof,
+    async checkStatus() {
+      return { verified: true }; // To work with latest version, we check status elsewhere
```
lovesh, line 236: This will make all credentials with invalid status pass? Why ignore status?

package.json
SamHellawell: debug, will remove, my bad

lovesh (approved):

PR #179, opened 2020-09-28
Fix jsonld-signature and jsonld dependency issues

SamHellawell:
also removes some old schema/credentials that don't parse as valid JSON-LD (assuming due to incorrect urls) causing examples to fail

### DIFF
--- a/example/standard_schemas.js
+++ b/example/standard_schemas.js
@@ -1,50 +1,9 @@
 import Schema from '../src/modules/schema';
 import { validateCredentialSchema, expandJSONLD } from '../src/utils/vc';
 
-import bolSchema from '../src/utils/vc/schemas/bol';
 import prCardSchema from '../src/utils/vc/schemas/pr_card';
-import qpInbonSchema from '../src/utils/vc/schemas/qp_inbond';
-import healthWorkerPassportSchema from '../src/utils/vc/schemas/health_worker_passport';
-import infectionDiagnosisSchema from '../src/utils/vc/schemas/infection_diagnosis';
 import immunityEventRecordSchema from '../src/utils/vc/schemas/immunity_event_record';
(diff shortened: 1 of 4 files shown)


example/standard_schemas.js
```diff
@@ -1,51 +1,10 @@
 import Schema from '../src/modules/schema';
 import { validateCredentialSchema, expandJSONLD } from '../src/utils/vc';
 
-import bolSchema from '../src/utils/vc/schemas/bol';
```
lovesh, line 4: Why are some of the schemas being removed?

src/utils/vc.js
```diff
@@ -144,7 +144,10 @@ export function checkCredentialContext(credential) {
 function hasDockRevocation(status) {
   const id = status[credentialIDField];
   if (status
-    && jsonld.getValues(status, credentialTypeField).includes(RevRegType)
+    && (
+      jsonld.getValues(status, credentialTypeField).includes(RevRegType)
+      || jsonld.getValues(status, credentialTypeField).includes(`/${RevRegType}`)
```
lovesh, line 149: What pattern is this regex used to match?

SamHellawell: this isnt regex

example/standard_schemas.js
SamHellawell: the credentials are failing to be parsed as valid JSON-LD because it uses URLs that don't exist

SamHellawell: atleast in the 2.0.2 version these fail

src/utils/vc.js
SamHellawell: to explain a bit more, in json-ld 2.0.2 this method will potentially return value with a slash `/` prepended, so we are checking for that. later json-ld lib doesn't do such a thing

PR #181, opened 2020-10-02
add sonarjs to eslint rules, address sonarjs complaints

SamHellawell:


src/verifiable-credential.js
```diff
@@ -303,6 +246,76 @@ class VerifiableCredential {
       schemaApi,
     });
   }
+
```
lovesh, line 249: Are the changes below related to Sonarjs's complains?

lovesh (approved):

PR #183, opened 2020-10-06
Claim deduction example script

bddap:
blocked by https://github.com/docknetwork/sdk/pull/169


SamHellawell:
you need to fix your branch history, would advise to just make new branch from master and cherry pick your latest commits

bddap:
> you need to fix your branch history, would advise to just make new branch from master and cherry pick your latest commits

This branch was based on the one in #169. The commit history was long because the this pr was created before #169 was merged. I just rebased onto current master so the commit history is fixed.

SamHellawell (approved):

PR #204, opened 2020-10-23
anchoring example

bddap:
This pr adds a reference example for posting PoE anchors, batched and otherwise.

It assumes the presence of the PoE anchoring module as implemented in https://github.com/docknetwork/dock-substrate/pull/89 .

~It's nearly finished just needs a couple more lines in the `check()` function to convert and return the result.~


bddap:
This is ready to review, but merging is blocked by https://github.com/docknetwork/dock-substrate/pull/89.

SamHellawell (approved):

PR #206, opened 2020-11-16
Bonus

lovesh:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@docknetwork/sdk",
-  "version": "0.3.7",
+  "version": "0.4.0",
```
SamHellawell, line 3: why such a big version change? will this PR break compatibility with other 0.3.7 or less versions? i doubt it, in which case it should be bumped to 0.3.8

SamHellawell (changes requested):

package.json
lovesh: Done

SamHellawell (approved):

PR #208, opened 2020-11-18
Anchoring docs

bddap:
This is ready for review ~but it branches of of `anchoring-example` so we should wait for #204 to be merged before merging this.~

~Commit cc035e1 is the only change.~

### DIFF
--- a/tutorials/src/SUMMARY.md
+++ b/tutorials/src/SUMMARY.md
@@ -6,6 +6,7 @@
   - [Verifiable credentials and presentations](concepts_vcdm.md)
   - [Schema](concepts_blobs_schemas.md)
   - [Claim Deduction](concepts_claim_deduction.md)
+  - [PoE Anchors](concepts_poe_anchors.md)
 - [Tutorials](tutorials.md)
   - [DID](tutorial_did.md)
   - [DID Resolver](tutorial_resolver.md)
@@ -13,3 +14,4 @@
   - [Revocation](tutorial_revocation.md)
   - [Schema](tutorial_blobs_schemas.md)
   - [Claim Deduction](tutorial_claim_deduction.md)
+  - [PoE Anchors](tutorial_poe_anchors.md)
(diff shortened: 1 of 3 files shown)


example/anchor.js
```diff
@@ -0,0 +1,172 @@
...
+//
+// It is possible simply post `hash(document)` as an anchor, but it's recommended to double-hash
+// the document instead. Since `hash(hash(document)) = compute_root(hash(document))` a double-hashed
+// anchor can be interpreted as the root of a merkle-tree with 1 leaf. Since the merkle tree has
+// only one leaf, the proof of inclusion for that leaf will be empty.
+
+/* eslint-disable import/no-extraneous-dependencies, import/newline-after-import */
+
+/* eslint-disable camelcase */
+import { compute_root, create_proof, verify_proof } from 'mrklt';
```
lovesh, line 29: Can you add the support for batching utils in the SDK source as examples are not available when SDK build is used? Also, can you add a single function to generate tree from given documents, and return root and proofs as thats how this is supposed to be used. Calling `create_proof` for each document (leaf) recreates the tree which becomes expensive when trying to anchor 1000s of documents.

As an enhancement, hash function (for tree) can be abstracted by accepting a hash function/hash interface as an arg. such that callers are not forced to use blake2s (for wasm as well).

example/anchor.js
```diff
@@ -0,0 +1,172 @@
...
+  const single = blake2s(randomAsU8a());
+  assert(await checkBatched(single, []) === null);
+  await anchorBatched([single]);
+  assert(await checkBatched(single, []) !== null);
+}
+
+// Post a value to the anchors module.
+async function anchor(hash) {
+  const nc = await conn;
+  await signExtrinsic(nc.tx.anchor.deploy(u8aToHex(hash)))
```
lovesh, line 64: Can you add an anchor module supporting read/write of anchors so caller don't need to use the low level api object like `api.tx.anchor` and `api.query.anchor.anchors` but can use `dock.anchor.write` and `dock.anchor.check` and the anchor can be passed as bytes or hex?

example/anchor.js
bddap: Added ticket for optimization: https://github.com/docknetwork/mrklt/issues/3
Added ticket for enhancement: https://github.com/docknetwork/mrklt/issues/4

example/anchor.js
bddap: This is possible and I can see it improving developer experience.

<details><summary>Side comment</summary>
I'm personally of the opinion that accepting arguments in multiple formats can be harmful.
</details>

The anchoring [rfc explicitly notes this as out-of-scope](https://github.com/docknetwork/planning/blob/master/rfc/0007-anchoring.md#non-goals). I think it should be a standalone task and not part of the milestone. Created ticket: https://github.com/docknetwork/sdk/issues/209

lovesh (approved):

PR #215, opened 2020-12-11
language tagged string equality

### LINKED ISSUE #173: Support language tagged strings in claimgraph representation.
https://www.w3.org/TR/rdf11-concepts/#section-Graph-Literal
https://github.com/docknetwork/sdk/blob/1009c20b5adc96010f7029123b746c53c59ba0e7/src/utils/claimgraph.js

bddap:
fixes #173

### DIFF
--- a/src/utils/claimgraph.js
+++ b/src/utils/claimgraph.js
@@ -5,6 +5,18 @@
 // a blank node: `{ Blank: '<string>' }`,
 // or a literal node: `{ Literal: { value: '<string>', datatype: '<iri>' } }`
 //
+// In compliance with https://www.w3.org/TR/rdf11-concepts/#section-Graph-Literal
+// a literal node must have a language tag if and only if datatype is
+// `http://www.w3.org/1999/02/22-rdf-syntax-ns#langString`
+//
+// ```
+// { Literal: {
+//    value: '<string>',
+//    datatype: 'http://www.w3.org/1999/02/22-rdf-syntax-ns#langString',
+//    language: '<tag>',
+// } }
+// ```
+//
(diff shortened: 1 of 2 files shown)


src/utils/claimgraph.js
```diff
@@ -23,14 +35,23 @@ export function fromJsonldjsCg(jcg) {
 
 // convert a single node from json-ld claimgraph representation to this module's representation.
 function fromJsonldjsNode(jn) {
+  let ret;
```
SamHellawell, line 38: `ret` declaration can go below as its not used until `case 'Literal'`

src/utils/claimgraph.js
```diff
@@ -23,14 +35,23 @@ export function fromJsonldjsCg(jcg) {
...
   switch (jn.termType) {
     case 'NamedNode':
       return { Iri: jn.value };
     case 'BlankNode':
       return { Blank: jn.value };
     case 'Literal':
       assert(jn.datatype.termType === 'NamedNode', 'The datatype of an RDF literal must be an IRI');
-      return { Literal: { value: jn.value, datatype: jn.datatype.value } };
+      ret = { Literal: { value: jn.value, datatype: jn.datatype.value } };
```
SamHellawell, line 46: you can use `const ret` here, as the reference of `ret` never changes (it remains that object, you are just mutating its values)

SamHellawell (approved):
lgtm just minor feedback

src/utils/claimgraph.js
bddap: I did that originally but eslint got upset. It doesn't like block-scoped variables to be declared inside switch statements.

PR #218, opened 2020-12-24
POA-2 API implementation and replication of node tests

SamHellawell:
NOTE: Tests pass individually, but not when ran sequentially. Needs some investigating.

Tests take quite a while, and need a custom node branch to run so they will fail in our CI. PR also adds proper error handling for the api `signAndSend` method


package.json
```diff
@@ -56,6 +56,7 @@
     "build": "rollup -c",
     "watch": "rollup -c -w",
     "test": "jest --verbose ./tests/unit",
+    "test-democracy": "NODE_ENV=production jest --verbose --runInBand ./tests/integration/democracy",
```
SamHellawell, line 59: this line is temporary

SamHellawell:
test are now passing individually but are failing as a group, @lovesh could you take a look if you get time to see if you have any ideas to rectify

PR #233, opened 2021-02-25
DID Attestations

SamHellawell:
Waiting on mainnet changes before we can merge this

### DIFF
--- a/src/modules/did.js
+++ b/src/modules/did.js
@@ -88,7 +88,7 @@ class DIDModule {
   /**
    * Creates an attestation claim on chain for a specific DID
    * @param {string} attester - Attester's DID
-   * @param {Attestation} attestation - Attestation bytes
+   * @param {object} attestation - Attestation object with priority and iri
    * @param {Signature} signature - Signature from existing key
    */
   async attestClaim(attester, attestation, signature, waitForFinalization = true, params = {}) {
--- a/src/modules/did.js
+++ b/src/modules/did.js
@@ -9,7 +9,7 @@ import { getStateChange } from '../utils/misc';
 
 import Signature from '../signatures/signature'; // eslint-disable-line
 
-export const ATTESTS_IRI = 'attests'; // TODO: decide on this before merge
+export const ATTESTS_IRI = 'https://rdf.dock.io/alpha/2021#attestsDocumentContents';
 
 /** Class to create, update and destroy DIDs */
 class DIDModule {
--- a/src/modules/did.js
+++ b/src/modules/did.js
@@ -9,7 +9,7 @@ import { getStateChange } from '../utils/misc';
 
 import Signature from '../signatures/signature'; // eslint-disable-line
 
-const ATTESTS_IRI = 'attests'; // TODO: decide on this before merge
+export const ATTESTS_IRI = 'attests'; // TODO: decide on this before merge
 
 /** Class to create, update and destroy DIDs */
 class DIDModule {
--- a/src/modules/did.js
+++ b/src/modules/did.js
@@ -1,4 +1,5 @@
 import { encodeAddress } from '@polkadot/util-crypto';
+import { u8aToString, hexToU8a } from '@polkadot/util';
 import b58 from 'bs58';
 
 import {
@@ -7,6 +8,8 @@ import {
 import { getStateChange } from '../utils/misc';
 
 import Signature from '../signatures/signature'; // eslint-disable-line
+
+const ATTESTS_IRI = 'attests'; // TODO: decide on this before merge
 
 /** Class to create, update and destroy DIDs */
 class DIDModule {
@@ -83,12 +86,34 @@ class DIDModule {
   }
 
   /**
+   * Creates an attestation claim on chain for a specific DID
+   * @param {string} attester - Attester's DID
+   * @param {Attestation} attestation - Attestation bytes
+   * @param {Signature} signature - Signature from existing key
+   */
+  async attestClaim(attester, attestation, signature, waitForFinalization = true, params = {}) {
+    const hexId = getHexIdentifierFromDID(attester);
+    const attestTx = this.api.tx.attest.setClaim(hexId, attestation, signature.toJSON());
+    return await this.signAndSend(attestTx, waitForFinalization, params);
+  }
+
+  /**
    * Create the fully qualified DID like "did:dock:..."
    * @param {string} did - DID
    * @return {string} The DID identifier.
    */
   getFullyQualifiedDID(did) {
     return `${DockDIDQualifier}${did}`;
+  }
+
+  /**
+   * Fetches the DIDs attestations IRI from the chain
+   * @param {string} hexId - DID in hex format
+   * @return {Promise<object>} The DID attestations
+   */
+  async getAttests(hexId) {
+    const attests = await this.api.query.attest.attestations(hexId);
+    return attests.iri.isSome ? u8aToString(hexToU8a(attests.iri.toString())) : '';
   }
 
   /**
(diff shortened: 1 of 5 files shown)


bddap:
TODO:
- [ ] change docker image tag in scripts/config to a fastblock version of master before merging

src/modules/did.js
```diff
@@ -82,6 +85,18 @@ class DIDModule {
     return await this.signAndSend(this.createRemoveTx(didRemoval, signature), waitForFinalization, params);
   }
 
+  /**
+   * Creates an attestation claim on chain for a specific DID
+   * @param {string} attester - Attester's DID
+   * @param {object} attestation - Attestation object with priority and iri
+   * @param {Signature} signature - Signature from existing key
+   */
+  async attestClaim(attester, attestation, signature, waitForFinalization = true, params = {}) {
```
bddap, line 94: claim and attestation mean the same thing. `attestClaim` = `attestAttest`. Maybe `setAttestedDocument` would be easier to understand.

src/modules/did.js
```diff
@@ -91,6 +106,16 @@ class DIDModule {
...
 
+  /**
+   * Fetches the DIDs attestations IRI from the chain
+   * @param {string} hexId - DID in hex format
+   * @return {Promise<object>} The DID attestations
+   */
+  async getAttests(hexId) {
+    const attests = await this.api.query.attest.attestations(hexId);
+    return attests.iri.isSome ? u8aToString(hexToU8a(attests.iri.toString())) : '';
+  }
```
bddap, line 117:
```suggestion
  /**
   * Fetches the DIDs attestations IRI from the chain
   * @param {string} hexId - DID in hex format
   * @return {Promise<string | null>} The DID's attestation, if any
   */
  async getAttests(hexId) {
    const attests = await this.api.query.attest.attestations(hexId);
    return attests.iri.isSome ? u8aToString(hexToU8a(attests.iri.toString())) : null;
  }
```

src/utils/did.js
```diff
@@ -185,3 +185,43 @@ export async function createSignedDidRemoval(didModule, did, currentKeyPair) {
   const signature = signDidRemoval(didModule, didRemoval, currentKeyPair);
   return [didRemoval, signature];
 }
+
+/**
+ * Create and return a `Attestation` as expected by the Substrate node. Signing is intentionally kept separate as the
+ * JS code may not have access to the signing key like in case of hardware wallet.
+ * @param {Number} priority - Attestation priority
+ * @param {*} iri - Attestation iri
```
bddap, line 193: Would this work? Users need to be able to set attestation iri to None on-chain in order to remove their previous attestation.

```suggestion
 * @param {string | null} iri - Attestation iri
```

This may be out of scope for this pr so maybe best to just create a ticket for it.

tests/integration/did.test.js
```diff
@@ -59,6 +60,22 @@ describe('DID Module', () => {
     expect(!!result).toBe(true);
   }, 10000);
 
+  test('Can attest with a DID', async () => {
```
bddap, line 63: A few more tests I think belong here:
- ensure attestations can be deleted by setting IrI on-chain to `None`
- for any `Attestation` a, `deserialize(serialize(a)) == a`
  - test the above where `priority > 2^8-1` (want to make sure compact encoding is handled properly)
- create did, create attestation, assert exact contents of did document

Those tests can be considered out of scope for this pr so we could create a new ticket to add them.

src/modules/did.js
```diff
@@ -91,6 +106,16 @@ class DIDModule {
...
   }
 
+  /**
+   * Fetches the DIDs attestations IRI from the chain
+   * @param {string} hexId - DID in hex format
+   * @return {Promise<object>} The DID attestations
+   */
+  async getAttests(hexId) {
+    const attests = await this.api.query.attest.attestations(hexId);
+    return attests.iri.isSome ? u8aToString(hexToU8a(attests.iri.toString())) : '';
```
bddap, line 116: fyi, the bytes on-chain aren't guaranteed to be valid utf8.

src/modules/did.js
```diff
@@ -8,6 +9,8 @@ import { getStateChange } from '../utils/misc';
 
 import Signature from '../signatures/signature'; // eslint-disable-line
 
+export const ATTESTS_IRI = 'attests'; // TODO: decide on this before merge
```
bddap, line 12:
```suggestion
export const ATTESTS_IRI = 'https://rdf.dock.io/alpha/2021#attestsDocumentContents';
```

src/modules/did.js
bddap: Other possibilities:
- `setAttestation`
- `setClaim`
- `setAttestsDocumentContents`

src/utils/did.js
```diff
@@ -185,3 +185,43 @@ export async function createSignedDidRemoval(didModule, did, currentKeyPair) {
...
+export function signDidAttestation(didModule, did, attestation, currentKeyPair) {
+  const serializedAttestation = didModule.getSerializedAttestation(did, attestation);
+  return getSignatureFromKeyringPair(currentKeyPair, serializedAttestation);
+}
+
+/**
+ * Create a `Attestation` as expected by the Substrate node and signs it. Return the `Attestation` object and the signature
+ * @param {module} didModule - The did module
+ * @param {Number} priority - Attestation priority
+ * @param {*} iri - Attestation iri
```
bddap, line 219: `*` -> `string | null` ?

bddap (changes requested):

bddap:
Minimum requested changes before approval.
- ATTESTATION_IRI
- https://github.com/docknetwork/sdk/blob/ce7513bb0a3ff3b52e328946ca9baf971f3b25eb/src/modules/did.js#L182

src/modules/did.js
```diff
@@ -147,14 +175,23 @@ class DIDModule {
     //   serviceEndpoint: 'https://dock.io/vc/'
     // }];
 
-    return {
+    // Construct document
+    const document = {
       '@context': 'https://www.w3.org/ns/did/v1',
       id,
+      attests,
```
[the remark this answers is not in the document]
SamHellawell: good spot, was from testing

src/modules/did.js
SamHellawell: yep fair enough, i'll go with `setClaim` to be like the substrate call. initial idea with this method naming was to differentiate as a api.tx.attests call

src/utils/did.js
SamHellawell: that should work i think, `any` type was left from when i was returning the object initially before our conversation

tests/integration/did.test.js
SamHellawell: yeah i was thinking on what tests may be beneficial here, often its been the case to replicate some node-side tests. i'd like to also handle some expected failure cases. however these tests are extra, providing they are covered on node id also agree theyre out of scope for this PR but we should have a story to write them just so we can get the functionality merged

src/modules/did.js
SamHellawell: are they guaranteed to always be strings (human readable or?) or could they be "random" bytes? if non-strings are expected, then i can remove the u8aToString conversion and just return the u8a from hex instead, moving that string check to the test. wdyt?

src/utils/did.js
SamHellawell: refer to comment above on expected data, perhaps string may not be returned here but done on user-side when using the result

src/modules/did.js
bddap: They are only meaningful as strings. If they are "random" bytes they are invalid. It is possible to post an invalid string on-chain because the chain doesn't care.

It does make sense to always attempt a conversion to string. Also makes sense to be conscious of the behavior of `u8aToString`. Maybe it throws an error? Maybe it returns a string no matter what (that could be bad?)? I don't know.

SamHellawell: It'll return a string no matter what given that a string is just a uint8 array anyway behind the scenes, it essentially will convert the Uint8Array type to String.

bddap:
Here is the tag for the fastblock build of the latest `dock-substrate` `master`.

`docknetwork/dock-substrate:5487f84e2f1c88f8fdd7464d35e0a94d2e65a18a-fastblock`

tests/integration/did.test.js
bddap: https://github.com/docknetwork/sdk/issues/242

bddap (approved):

PR #231, opened 2021-02-22
Quads

bddap:
The PR modifies sdk to use the latest version of rify, which can reason over RDF quads.

- [x] Refactor cg and claimgraph modules to operate on quads.
- [x] Remove Explicit Ethos reification
- [x] Modify unit test rulesets to express attestation logic as quads instead of in reified EE
- [x] refactor unit tests
- [x] refactor affected examples
- [x] refactor affected tutorials
- [x] rebase on master after #229 is merged
- [x] use the latest non-release-candidate version of rify after https://github.com/docknetwork/rify/pull/10 is merged (using 0.6.0-rc.1 right now)
- [x] proofread claim deduction concepts doc again before submitting for review

PR #234, opened 2021-03-03
Private delegation logic

bddap:
Blocked by #231

Don't merge this directly. The target of this pr is `quads` but it should be changed to `master` before merging.

### DIFF
--- a/src/rdf-defs.js
+++ b/src/rdf-defs.js
@@ -1,0 +1,165 @@
+// @prefix dockalpha: <https://rdf.dock.io/alpha/2021#>
+
+// id: dockalpha:mayClaim
+// type: rdf:Property
+// domain: rdf:Resource
+// range: dockalpha:ClaimFilter
+//
+// An entity that attests to a dockalpha:mayClaim statement grant the subject of that statement
+// authority to make any claims statisfying the object of the statment on the entities behalf.
+export const MAYCLAIM = 'https://rdf.dock.io/alpha/2021#mayClaim';
+
+// id: dockalpha:ANYCLAIM
+// type: dockalpha:ClaimFilter
+//
+// The "allow all" filter. dockalpha:ANYCLAIM accepts all triples.
+// All triples satisfy dockalpha:ANYCLAIM.
+export const ANYCLAIM = 'https://rdf.dock.io/alpha/2021#ANYCLAIM';
+
+// id: dockalpha:ClaimFilter
+// type: rdf:Class
+//
+// Filter over rdf claims.
+export const CLAIM_FILTER = 'https://rdf.dock.io/alpha/2021#ClaimFilter';
+
+// This ruleset defines dockalpha:MayClaim specifically only when used with dockalpha:ANYCLAIM.
+//
+// This is not a complete definition of dockalpha:MayClaim, but any complete definition of
+// dockalpha:MayClaim will imply this ruleset.
+//
+// This ruleset is generic over graph name. The genericity of the rule allows it to prove composite
+// claims where graph != the default graph, but at some potential reasoning cost. See
+// MAYCLAIM_DEF_2 for a lighter-wheight partial definition.
+export const MAYCLAIM_DEF_1 = [
+  {
+    if_all: [
+      [
+        { Unbound: 'b' },
+        { Bound: { Iri: MAYCLAIM } },
+        { Bound: { Iri: ANYCLAIM } },
+        { Unbound: 'a' },
+      ],
+      [
+        { Unbound: 's' },
+        { Unbound: 'p' },
+        { Unbound: 'o' },
+        { Unbound: 'b' },
+      ]
+    ],
+    then: [
+      [
+        { Unbound: 's' },
+        { Unbound: 'p' },
+        { Unbound: 'o' },
+        { Unbound: 'a' },
+      ]
+    ],
+  }
+];
+
+// This ruleset defines dockalpha:MayClaim specifically only when used with dockalpha:ANYCLAIM.
+//
+// During reasoning this ruleset is more efficient than its predecessor in terms of space used.
+// This rule generates fewer tuples becuse claims are only ever copied into the default graph.
+// When using this rule tuples need not propagate from graph to graph all the way up the delegation
+// chain. Rather, when it's determined that a specific issuer is authorized, claims will be copied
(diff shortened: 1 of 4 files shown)


tests/cached-document-loader.js
```diff
@@ -0,0 +1,46 @@
+import contexts from '../src/utils/vc/contexts';
```
bddap, line 1: The test for private delegation required a document loader. I pulled the `documentLoader` out of `claim-deduction.test.js`, so it could be used by the other tests.

tests/cached-document-loader.js
```diff
@@ -0,0 +1,46 @@
...
+export function addDocument(iri, doc) {
+  documentRegistry[iri] = defensive_copy(doc);
+}
+
+export async function modifyDocument(iri, cb) {
+  let original = (await documentLoader(iri)).document;
+  addDocument(iri, cb(original));
+}
+
+/// add a document to the registry
```
bddap, line 39:
```suggestion
/// check if document is already cached/registered
```

tests/unit/claim-deduction.test.js
```diff
@@ -1,6 +1,5 @@
 import { Ed25519KeyPair, suites } from 'jsonld-signatures';
```
bddap, line 1: The changes in this document are only because I pulled the `documentLoader` out of this file and put it in `tests/cached-document-loader.js`.

tests/unit/logic.test.js
```diff
@@ -0,0 +1,126 @@
+import { presentationToEEClaimGraph, proveh } from '../../src/utils/cd';
+import jsonld from 'jsonld';
+import { documentLoader } from '../cached-document-loader.js';
+import { randomAsHex } from '@polkadot/util-crypto';
+import { ANYCLAIM, MAYCLAIM, MAYCLAIM_DEF_1, MAYCLAIM_DEF_2 } from '../../src/rdf-defs';
+
+const COOLNESS = 'https://example.com/coolnessLevel';
+
+describe('logic', () => {
+  test('delegation chain', async () => {
```
bddap, line 10: This test shows that when a delegation chain is bundled in a presentation along with a credential, `MAYCLAIM_DEF_1`, `MAYCLAIM_DEF_2` can be used to prove a attested statement.

This test also checks to make sure removing parts of the delegation chain, making it invalid, makes the target statement unprovable.

tests/unit/logic.test.js
```diff
@@ -0,0 +1,126 @@
...
+    };
+
+    let to_prove = [[
+      { Iri: 'did:example:d' },
+      { Iri: COOLNESS },
+      { Literal: { datatype: 'http://www.w3.org/2001/XMLSchema#integer', value: '1' } },
+      { DefaultGraph: true },
+    ]];
+
+    let assumed = [[
```
bddap, line 51:
```suggestion
    // Manually grant 'did:example:a' authority to make claims. This makes 'did:example:a' a root authority.
    let assumed = [[
```

tests/unit/logic.test.js
```diff
@@ -0,0 +1,126 @@
...
+    // happy path
+    cg = cat(await asCg([cred, del1, del2]), assumed);
+    proveh(cg, to_prove, MAYCLAIM_DEF_1);
+    proveh(cg, to_prove, MAYCLAIM_DEF_2);
+
+    // the assumption assigning a root authority is not included in the claimgraph
+    cg = await asCg([cred, del1, del2]);
+    assert_error(
+      () => proveh(cg, to_prove, MAYCLAIM_DEF_1),
+      { CantProve: "NovelName" },
```
bddap, line 69: Normally an unprovable composite claim will result in the error: `{ CantProve: "ExhaustedSearchSpace" }` but not in this specific case.

`{ DefaultGraph: true }` is never mentioned in the input `cg`. Neither is it mentioned in the output clause of `MAYCLAIM_DEF_1`. Rify will (correctly) infer that it is impossible to prove a composite claim containing `{ DefaultGraph: true }` so it will skip reasoning entirely.

https://github.com/docknetwork/rify/blob/ac12e6ad1fc23ac232e3d256e38edf9ad0ee69f7/src/prove.rs#L191

tests/unit/logic.test.js
```diff
@@ -0,0 +1,126 @@
...
+      "https://www.w3.org/2018/credentials/v1"
+    ],
+    '@type': ['VerifiablePresentation'],
+    'id': `urn:${randomAsHex(16)}`,
+    'holder': `urn:${randomAsHex(16)}`,
+    'verifiableCredential': creds,
+  };
+}
+
+function cat(...arrays) {
```
bddap, line 120: `cat([1, 2, 3], [4, 5, 6], [7]) = [1, 2, 3, 4, 5, 6, 7]`

src/rdf-defs.js
```diff
@@ -0,0 +1,171 @@
...
+//
+// The "allow all" filter. dockalpha:ANYCLAIM accepts all triples.
+// All triples satisfy dockalpha:ANYCLAIM.
+export const ANYCLAIM = 'https://rdf.dock.io/alpha/2021#ANYCLAIM';
+
+// id: dockalpha:ClaimFilter
+// type: rdf:Class
+//
+// Filter over rdf claims.
+export const CLAIM_FILTER = 'https://rdf.dock.io/alpha/2021#ClaimFilter';
```
bddap, line 24: This section is the start of an ontology definition like https://www.w3.org/1999/02/22-rdf-syntax-ns# .

src/rdf-defs.js
```diff
@@ -0,0 +1,171 @@
...
+//         { Unbound: 'b' },
+//         { Bound: { Iri: MAYCLAIM } },
+//         { Bound: { Iri: ANYCLAIM } },
+//         { Unbound: 'a' },
+//       ],
+//     ],
+//   }
+// ];
+
+// # The following considerations, and more, are why these rules will be marked as alpha.
```
bddap, line 134:
```suggestion
// # The following considerations, and more, are why these definitions (e.g. dockalpha:mayClaim) are marked as alpha.
```

nmrshll (approved):
I've had a look at the tests in logic.test.js and claim-deduction.test.js,
they make sense and I can't see any obvious one to add, so looks good to merge to me !

PR #236, opened 2021-03-09
Private delegation concepts and tutorial

bddap:
rendered
- https://github.com/docknetwork/sdk/blob/private-delegation-tuts/tutorials/src/concepts_private_delegation.md
- https://github.com/docknetwork/sdk/blob/private-delegation-tuts/tutorials/src/tutorial_private_delegation.md

### DIFF
--- a/tutorials/src/tutorial_private_delegation.md
+++ b/tutorials/src/tutorial_private_delegation.md
@@ -1,3 +1,114 @@
 # Private Delegation
 
-TODO
+This tutorial follows the lifecycle of a delegated credential. It builds builds on previous tutorials. Namely, this tutorial assumes you have already grokked [Issuance, Presentation, Verification](./tutorial_ipv.md) and [Claim Deduction](./tutorial_claim_deduction.md).
+
+## Create a Delegation
+
+<details>
+<summary>Boilerplate</summary>
+
+```js
+const { v4: uuidv4 } = require('uuid');
+
+function uuid() {
+  return `uuid:${uuidv4()}`;
+}
+
+// Check out the Issuance, Presentation, Verification tutorial for info on signing
+// credentials.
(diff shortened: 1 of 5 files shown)


tutorials/src/tutorial_private_delegation.md
```diff
@@ -0,0 +1,116 @@
...
+```
+
+`presentation` is sent to the verifier.
+
+## Accept a Delegated Credential
+
+The verifier receives `presentation`, *verifies the enclosed credentials*, then reasons over the union of all the credentials in the bundle using the rule `MAYCLAIM_DEF_1`. The process is the one outlined in [Verifier-Side Reasoning](./tutorial_claim_deduction.md#verifier-side-reasoning) but using a different composite claim and a different rule list.
+
+```js
+import { MAYCLAIM_DEF_1 } from '../src/rdf-defs';
```
nmrshll, line 92: For users of the library this is going to be imported from `@docknetwork/sdk`.
Should we update the examples to import from there (rather than `../`),
 so that users can copy-paste the code and get it working faster ?

SamHellawell: Agree here, although its difficult to update the actual example code in this repo, for the markdown documents we should def replace with `@docknetwork/sdk`

SamHellawell (approved):
lgtm besides @nmrshll's point of feedback

tutorials/src/tutorial_private_delegation.md
bddap: @SamHellawell , would this be valid?

```suggestion
import { MAYCLAIM_DEF_1 } from '@docknetwork/sdk/rdf-defs';
```

SamHellawell: yeah that looks fine

PR #239, opened 2021-03-15
RDF Turtle parsing + IPFS storage

SamHellawell:


src/utils/rdf.js
```diff
@@ -0,0 +1,83 @@
...
+  return result.join('\n');
+}
+
+/**
+ * Parses an RDF document string and formats it according to rify requirements
+ * @param {string} document - RDF document in turtle or other form
+ * @param {object} parserOptions - N3 Parser configuration object
+ * @returns {*}
+ */
+export function parseRDFDocument(document, parserOptions = {}) {
```
bddap, line 50: When crawling we should be strict about the type of document we accept. I think we need a function that will accept only turtle.

My primary concern is that the meaning of a document may change depending on the format a parser assumes. I don't want it to be possible for attestations have multiple interpretations.

Is there are reason this should accept multiple formats?

bddap: Passing `{ format: "text/turtle" }` to the parser constructor will force turtle.

tests/unit/rdf-turtle.test.js
```diff
@@ -0,0 +1,112 @@
...
+
+  test('Cannot parse RDF document with non-default graph', async () => {
+    expect(() => {
+      parseRDFDocument(rdfInputs[2]);
+    }).toThrowError(/Unexpected graph/);
+  });
+
+  // test('Can fetch RDF document from IPFS', async () => {
+  //   expect(true).toBe(false);
+  // });
```
bddap, line 111: Dead code.
```suggestion
```

src/utils/rdf.js
```diff
@@ -0,0 +1,83 @@
...
+
+let ipfsClient;
+
+/**
+ * Writes a document to IPFS and returns its CID
+ * @param {string} document - RDF document to store on IPFS
+ * @param {object} connectionConfig - IPFS HTTP Client connection options
+ * @returns {Promise<any>}
+ */
+export async function writeToIPFS(document, connectionConfig = ipfsDefaultConfig) {
```
bddap, line 15: Consider this code:

```js
try {
  await writeToIPFS('yeah', 'http://localhost:5005');
} catch {
  await writeToIPFS('yeah', 'http://localhost:5001');
}
```

tests/integration/ipfs.test.js
```diff
@@ -0,0 +1,20 @@
...
+
+const rdfCID = 'QmQeskBk9TsdWXaNX26PvuzkxU8zzoWZo5dmJMPuw5vcSS';
+
+describe('IPFS', () => {
+  test('Can write document to IPFs', async () => {
+    const cid = await writeToIPFS(rdfInput);
+    expect(cid.toString()).toEqual(rdfCID);
+  }, 10000);
+
+  test('Can dereference document from IPFs', async () => {
```
bddap, line 16: Test 2 implicitly depends on test 1.
```suggestion

```

tests/integration/ipfs.test.js
```diff
@@ -0,0 +1,20 @@
...
+    const document = await dereferenceFromIPFS(rdfCID);
+    expect(document).toEqual(rdfInput);
+  }, 10000);
+});
```
bddap, line 20: Is it possible to add a test to prove that dereferencing a live *ipns* document results in an error?

bddap (commented):
I'm concerned with parseRDF being non-strict about the document syntax it accepts.

My note about the ipfs functions is less important.

src/utils/rdf.js
SamHellawell: That's what parseroptions are for when you call this method, this method shouldn't force turtle as it may be used not by crawler since we're offering it in the SDK

src/utils/rdf.js
SamHellawell: do you think that user will want to connect to multiple nodes in one program?

tests/unit/rdf-turtle.test.js
SamHellawell: good spot, ty

tests/integration/ipfs.test.js
SamHellawell: yes it does, and i don't see how your suggestion improves that or why its an issue. your suggestion only makes it harder to see what goes wrong, imo

tests/integration/ipfs.test.js
SamHellawell: why would dereferencing a live ipns document result in an error?

src/utils/rdf.js
SamHellawell: you know, i dont think its wise actually to include `ipfs-http-client` module in SDK and providing these very very simple methods to write for example. i think dereference could be useful if we pass the client from the caller, rather than in the SDK

tests/integration/ipfs.test.js
SamHellawell: was able to reduce the tests by remove write method, assuming that is now covered by ipfs-http-client library tests

tests/integration/ipfs.test.js
bddap: Ipns docs are mutable. This prevents a verifier from knowing when to invalidate data that was pulled derived from that sort of document.

Let's say the crawler extracts a claimgraph from an ipns document. There isn't a straightforward way for the verifier to know that the document changed and that a re-crawl is required.

bddap (approved):
Thank you Sam for figuring these things out.

src/utils/rdf.js
bddap: I see. As long as the crawler has the ability to force triples it should be safe.

PR #240, opened 2021-03-18
SPARQL claimgraph querying

SamHellawell:


src/utils/rdf.js
```diff
@@ -1,6 +1,75 @@
...
+      toJsonldjsNode(object),
+      graph,
+    ));
+  });
+  return store;
+}
+
+/**
+ * Queries a claimgraph object and returns its bindings as an RDF term list
+ * @param {object} claimgraph - JSON-LD RDF object
```
bddap, line 52:
```suggestion
 * @param {object} claimgraph - A list of RDF quads
```

src/utils/rdf.js
```diff
@@ -1,6 +1,75 @@
...
+ * @param {string} query - SPARQL query string
+ * @param {object|null} engine - RDF query engine or null to auto-create
+ * @returns {array<any>}
+ */
+export async function queryClaimgraph(claimgraph, query, engine = null) {
+  // Create an N3 store from claimgraph JSON object
+  const store = typeof claimgraph === 'string' ? claimgraph : claimgraphToStore(claimgraph);
```
bddap, line 59: nit: Arguments that can be one of a number of types makes me uncomfortable, especially when the behavior is not documented.

bddap: nit: there doesn't seem to be any test that exercises the ability to pass claimgraph as a string.

tests/unit/rdf-turtle.test.js
```diff
@@ -24,6 +24,63 @@ const rdfInputs = [
...
+  test('Can query a claimgraph', async () => {
+    const query = `prefix dockalpha: <http://rdf.dock.io/alpha/2021#>
+    select * {
+      ?root dockalpha:mayClaim dockalpha:ANYCLAIM .
+      graph ?root {
+        ?lookupNext dockalpha:mayClaim dockalpha:ANYCLAIM .
+      }
+    }`;
+    const newGraph = await queryClaimgraph(claimgraph, query);
+    expect(newGraph).toEqual([
```
bddap, line 78:
```suggestion
    const queryResult = await queryClaimgraph(claimgraph, query);
    expect(queryResult).toEqual([
```

src/utils/rdf.js
```diff
@@ -1,6 +1,75 @@
...
+  // Query the engine, if querying multiple times user should
+  // pass engine parameter for optimal performance
+  const myEngine = engine || newEngine();
+  const result = await myEngine.query(query, { sources: [store] });
+
+  // Get bindings from query
+  const bindings = await result.bindings();
+
+  // Convert bindings to claimgraph format using lookupNext variable
+  return bindings.map((binding) => fromJsonldjsNode(binding.get('?lookupNext')));
```
bddap, line 70: How will this behave for a query that is invalid because it has no the `?lookupNext` variable.

src/utils/rdf.js
```diff
@@ -1,6 +1,75 @@
...
+      toJsonldjsNode(predicate),
+      toJsonldjsNode(object),
+      graph,
+    ));
+  });
+  return store;
+}
+
+/**
+ * Queries a claimgraph object and returns its bindings as an RDF term list
```
bddap, line 51:
```suggestion
 * Queries a claimgraph object and returns all the bindings to the variable named `?lookupNext`
```

src/utils/rdf.js
```diff
@@ -1,6 +1,75 @@
...
+ * @param {object} claimgraph - JSON-LD RDF object
+ * @param {string} query - SPARQL query string
+ * @param {object|null} engine - RDF query engine or null to auto-create
+ * @returns {array<any>}
+ */
+export async function queryClaimgraph(claimgraph, query, engine = null) {
```
bddap, line 57:
```suggestion
export async function queryNextLookup(claimgraph, query, engine = null) {
```

bddap (approved):

src/utils/rdf.js
SamHellawell: I was thinking about this, and I think it would return potentially null/undefined in the array. I wasn't sure if its good to hardcode this, maybe exposing a parameter with `?lookupNext` as a default value would make more sense. That way a user can choose to get their own bindings if they wish, rather than this method just resulting in an error if not found

src/utils/rdf.js
SamHellawell: > nit: there doesn't seem to be any test that exercises the ability to pass claimgraph as a string.

wasn't aware that this was a required use case. what benefit is there to having this method support multiple types? i'd prefer that the string->obj conversion was done out of band. and by string, do you mean JSON string or some other format?

bddap: It's not a required use case. I don't think there is a benefit to the method supporting multiple types. I also prefer that the string->obj conversion happen out of band.

This line is checking whether claimgraph is a string. Why is that?

src/utils/rdf.js
bddap: As a reminder, this method doesn't need to be exposed to users. It's only needed for use by the crawler.

Users will be able to provide the sparql query, but they don't need to call this function directly.

bddap: To avoid hard-coding `lookupNext` this function could just return `bindings` and let the crawler do the rest.

src/utils/rdf.js
SamHellawell: my bad, i misunderstood your first comments. so, this was initially intended to provide support for URI sources. you're right in that its not entirely clear from the method. this was done when first testing queries, i don't think it's actually needed to be included here - but that depends on if you think the crawler would need URI sources. my assumption is no.

src/utils/rdf.js
SamHellawell: if we're just returning bindings then this method to me would seem useless, as its basically just a wrapper around query and bindings return. do you have anything against using a parameter to define which binding to get? im happy with `lookupNext` as a default value

SamHellawell: alternatively, we can just assume user has to use lookupNext in their query and then if we cant find any binding with that value defined we can throw an error in this method. it's hard to determine the best approach with no crawler implementation, feels a bit silo'd

src/utils/rdf.js
SamHellawell: i have removed it, ill mark as resolved - lmk if you feel like its useful supporting more store types

src/utils/rdf.js
bddap: I think any of  the proposed solutions works, including how it is now.

Thanks for mentioning that this feels silo'd. I think you are right and it's probably a pain to have to bounce stuff off me instead of me explaining my plans in more detail up-front.

PR #247, opened 2021-03-30
Add crawler for delegation v2

bddap:
TODO before merge:
- [x] Ensure graph hygiene.The crawler needs to rename or randomize blank nodes when merging new documents into it's knowledge graph.
- [x] Change rify version from `0.7.0-rc.1` to `0.7.0` once that version is released.


src/utils/claimgraph.js
```diff
@@ -69,7 +69,7 @@ export function fromJsonldjsNode(jn) {
 // Blank node name generator
 // An object of this type won't produce the same name twice, even accross multiple calls to
 // reallocateNames.
-class Namer {
+export class Namer {
```
bddap, line 72: @SamHellawell is there a good way to expose this to other parts of the sdk without exposing it to users?

bddap:
Putting this here for future reference please ignore
outfile

src/utils/claimgraph.js
SamHellawell: not really, by the nature of JS if its exported its publicly available through importing that file and that the SDK separates files. If we were to bundle a bunch of the files then sure, maybe it could be "hidden". Still, I don't think its a big concern. If you don't want to expose it in utils/claimgraph put it in another file and import it into claimgraph. That file will still be available for the user to import directly from if they want though.

src/crawl.js
```diff
@@ -0,0 +1,107 @@
+/*
+eslint no-await-in-loop: "off",
```
SamHellawell, line 2: this is on for a reason. you can keep it disabled if you really like, but it will likely hurt performance because you cant benefit from parallelization

src/crawl.js
```diff
@@ -0,0 +1,107 @@
...
+    // Add term to lookedup.
+    // Return whether the term was already present in the set.
+    // like https://doc.rust-lang.org/std/collections/btree_set/struct.BTreeSet.html#method.insert
+    const str = canon(term);
+    const isnew = !lookedup.has(str);
+    lookedup.add(str);
+    return isnew;
+  };
+
+  while (true) {
```
SamHellawell, line 40: i think this could be refactored to use promises and recursion, then you dont need to disable await in infinite loop for this

src/crawl.js
```diff
@@ -0,0 +1,107 @@
...
+    // reason over what we have so far
+    facts = facts.concat(inferh(facts, rules));
+
+    // lookup any interesting documents
+    const interesting = await queryNextLookup(facts, curiosityQuery);
+    const novel = interesting.filter(marknew);
+    const newfacts = [...await Promise.all(novel.map(resolveGraph))];
```
SamHellawell, line 47: why use spread syntax to create a new array? existing array returned by `await Promise.all(novel.map(resolveGraph))` should suffice

src/utils/claimgraph.js
bddap: Thanks

src/crawl.js
bddap: Fixed 915cbbcd40c4195b3c092ad42f6861b1fdaa2b48

src/crawl.js
bddap: Care to go into more detail?

src/utils/cd.js
```diff
@@ -217,6 +217,16 @@ export function validateh(rules, proof) {
   };
 }
 
+// A wrapper around infer that first converts premises and rules to the canonical representation
+// as defined by `canon()` in `claimgraph.js`. This wrapper deserializes the returned values before
+// passing them back to the caller.
+export function inferh(premises, rules) {
```
SamHellawell, line 223: not sure on the naming of this method, `inferh` isnt verify descriptive. what does the `h` stand for, if anything?

tests/unit/ls
```diff
@@ -0,0 +1,146 @@
+import crawl from '../../src/crawl.js';
```
SamHellawell, line 1: redundant file?

SamHellawell (approved):
couple of comments but lgtm for most part

src/utils/cd.js
bddap: I can add a comment stating that it's short for infer-high because it's a higher level wrapper around `infer`.
The same is true for `proveh` and `validateh`.

tests/unit/ls
bddap: Oops. This is one of the things I should have caught during self-review. Will delete.

bddap: removed a814beb

src/utils/cd.js
SamHellawell: i think we can just leave as is, dont think its something user would care about anyway. h for high level wrapper can do

bddap: Too late. :)
c12a16b

src/crawl.js
bddap: I just remembered why I did that. Its because I need to check the length of `newfacts`. [Promse.all](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) returns an `iterator` but that iterator is not guaranteed to be an array.

bddap: I could loop over the iterable to count it's elements but I think I should just change it back. I'm optimizing for readability here.

bddap: Reverted 915cbbc in f73c833 . Thanks for pointing it out though.

src/crawl.js
bddap: For context, I did already put some effort into de-looping (note use of Promise.all). The barrier I ran into is that this process is inherently sequential. The decision of which iris to lookup next depends on all previous results.

SamHellawell: i think we can overlook that general rule and keep as is for now, thinking about how to deloop it - its doable but it may just add un-necessary complication.

bddap:
Tests are passing but the CI is still failing. https://github.com/docknetwork/sdk/runs/2326871315#step:8:195

bddap:
So f73c833 passes but f507f69 does not. I think this is a problem with CI and not with these changes. I will rerun CI for f73c833 to confirm.

bddap:
Thanks for fixing CI @maycon-mello .

PR #253, opened 2021-05-04
Remove duplicate ethr-did-resolver dep

SamHellawell:


SamHellawell:
it seems some tests are flaky after polkadotjs upgrade, have the disconnection error that we got before

SamHellawell:
Closing PR as will be addressed in tech debt updates

PR #277, opened 2021-10-22
upgrade packages

maycon-mello:


SamHellawell:
resolved the issue with unit tests, now theres another issue with general issuing/crypto logic. polkadot's tohex methods are having problems recognising plain arrays

PR #280, opened 2021-11-09
Bbs plus and accumulator

lovesh:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@docknetwork/sdk",
-  "version": "1.0.2",
+  "version": "1.1.0",
```
SamHellawell, line 3: please update version as i pushed a node-types change that is version 1.1.0

src/index.js
```diff
@@ -100,6 +103,8 @@ class DockAPI {
     // Initialize price feed rpc
     rpc = Object.assign(rpc, PriceFeedRpcDefs);
 
+    rpc = Object.assign(rpc, CoreModsRpcDefs);
```
SamHellawell, line 106: we can do:
```
rpc = Object.assign(Object.assign(rpc, PriceFeedRpcDefs), CoreModsRpcDefs);
```

src/modules/WithParamsAndPublicKeys.js
```diff
@@ -0,0 +1,357 @@
...
+    } else if (curveType === 'Bls12381') {
+      publicKey.curveType = curveType;
+    } else {
+      throw new Error(`Invalid curve type ${curveType}`);
+    }
+    if (paramsRef !== undefined) {
+      // eslint-disable-next-line no-param-reassign
+      paramsRef = WithParamsAndPublicKeys.parseRef(paramsRef);
+    }
+    publicKey.params_ref = paramsRef;
```
SamHellawell, line 58: we can do:
```
publicKey.params_ref = paramsRef !== undefined ? WithParamsAndPublicKeys.parseRef(paramsRef) : undefined;
```

src/modules/WithParamsAndPublicKeys.js
```diff
@@ -0,0 +1,357 @@
...
+
+  /**
+   * Get all params written by a DID
+   * @param did
+   * @returns {Promise<object[]>}
+   */
+  async getAllParamsByDid(did) {
+    const hexId = getHexIdentifierFromDID(did);
+
+    // TODO: Figure out why this doesn't work
```
SamHellawell, line 218: is this comment still relevant?

SamHellawell (commented):
looks good for the most part, couple minor changes suggested and comments

it would be good to have an example file for this too, ideally it could show how an actual credential can be used with this. doesnt have to be in this PR though

package.json
lovesh: Done

lovesh:
> looks good for the most part, couple minor changes suggested and comments
> 
> it would be good to have an example file for this too, ideally it could show how an actual credential can be used with this. doesnt have to be in this PR though

Yes, i can later copy part of [this test](https://github.com/docknetwork/sdk/pull/280/files#diff-96d0480e32a3f04d4b6efe391b5fe10f836aec1d4f6eb7307b11064380debc8dR1) to the example.

src/index.js
lovesh: Is there an advantage to doing this. Because i like them on separate lines for readability. Added a comment btw

src/modules/WithParamsAndPublicKeys.js
lovesh: Good point. Done

src/modules/WithParamsAndPublicKeys.js
lovesh: Its relevant but not important now. I can do this optimization later

src/index.js
SamHellawell: personal preference, can keep it either way

SamHellawell (approved):

PR #284, opened 2022-01-03
Detect if http network and use appropriate provider

SamHellawell:


src/index.js
```diff
@@ -113,8 +120,11 @@ class DockAPI {
       rpc = Object.assign(rpc, PoaRpcDefs);
     }
 
+    const isWebsocket = this.address && this.address.indexOf('http') === -1;
```
mike-parkhill, line 119: could this be more specific? Or do a check above to make sure one of `ws` or `http` are being specified? I'd assume we should be failing on anything else, right?

SamHellawell: not checking for ws to be specified because i believe its possible to provide straight IP address (no ws/wss protocol prefix) and it should connect. so for it to be a websocket connection, we need the address not to contain http

mike-parkhill: Got it.

mike-parkhill (approved):

PR #286, opened 2022-01-17
Staking payouts script

olegnn:
Implementation of the staking payouts.
This script is intended to be triggered by the cron job or some other task scheduler.


scripts/staking_payouts.js
```diff
@@ -0,0 +1,367 @@
...
+/**
+ * Retuns era rewards need to be paid grouped by stash ids.
+ *
+ * @param {*} api
+ * @param {ErasInfo} erasInfo
+ * @param {*} validators
+ * @returns {Array<StashEras>}
+ */
+const getUnclaimedStashEras = (api, { eras }, validators) => {
+  const singleStashEras = async (api, stashAddress, allEras) => {
```
lovesh, line 254: This is supposed to return unclaimed eras for the stash, correct? Can you name/document this function appropriately?

scripts/staking_payouts.js
```diff
@@ -0,0 +1,367 @@
...
+    ).unwrapOrDefault();
+
+    return differenceWith((a, b) => a.eq(b), allEras, ledger.claimedRewards);
+  };
+
+  return from(validators).pipe(
+    mergeMap(
+      (stashAddress) =>
+        from(singleStashEras(api, stashAddress, eras)).pipe(
+          filterRx(complement(isEmpty)),
```
lovesh, line 266: `filterRx(complement` can be replaced by `reject`?

scripts/staking_payouts.js
```diff
@@ -0,0 +1,367 @@
...
+const getUnclaimedStashEras = (api, { eras }, validators) => {
+  const singleStashEras = async (api, stashAddress, allEras) => {
+    const ledger = (
+      await api.query.staking.ledger(stashAddress)
+    ).unwrapOrDefault();
+
+    return differenceWith((a, b) => a.eq(b), allEras, ledger.claimedRewards);
+  };
+
+  return from(validators).pipe(
```
lovesh, line 262: Can you also put a comment here about whats happening in the return?

scripts/staking_payouts.js
```diff
@@ -0,0 +1,367 @@
...
+ * Signs and sends extrinsics produced by the given observable.
+ *
+ * @param {*} api
+ * @param {*} initiator
+ * @param {Rx.Observable<import("@polkadot/types/interfaces").Extrinsic>} payouts$
+ *
+ * @returns {Rx.Observable<string>} executed tx with hashes
+ */
+const signAndSendPayouts = curry((api, initiator, payouts$) => {
+  const nonces$ = timer(0, 1000).pipe(
```
lovesh, line 284: I think you are trying to get few nonces, 1 for each batch and then send. But can you please document the sections appropriately.

scripts/staking_payouts.js
```diff
@@ -0,0 +1,367 @@
...
+            .lt(MaxCommission)
+        ) {
+          const reward = eraPoints.validators[stashAddress]
+            .mul(eraRewards.eraReward)
+            .div(eraPoints.eraPoints);
+
+          if (!reward.isZero())
+            return acc.concat({
+              era,
+              eraReward: eraRewards.eraReward,
```
lovesh, line 351: Do you need `eraReward` and `reward` (apart from debugging) as you are not showing this data anyway?

lovesh (changes requested):

scripts/staking_payouts.js
olegnn: Unfortunately, `rxjs` doesn't have a `reject` operator, the one in the code is the `Ramda`'s function.

scripts/staking_payouts.js
olegnn: `reward` for each account will be shown in logs, see https://github.com/docknetwork/sdk/pull/286/files#diff-4cf85c6f9e9ab3caf77a6239964ea82cf0e8c404e7abe067e38bc8468f108dcbR167
`eraReward` was removed as it's currently unused.

lovesh: I had seen the logs but I was saying if you could remove the reward as its not needed. Anyway, its just a script so efficiency doesn't matter much so if it helps to keep it then lets keep it.

scripts/email_utils.js
```diff
@@ -0,0 +1,38 @@
...
+export async function sendAlarmEmail(subject, body) {
+  const ses = new SESV2({
+    apiVersion: "2019-09-27",
+    accessKeyId: process.env.AwsAccessId,
+    secretAccessKey: process.env.AwsSecretKey,
+    region: process.env.AwsSesEmailRegion,
+  });
+
+  // Split comma separated list of email recipients
+  const toAddr = process.env.AlarmEmailTo.split(",");
```
lovesh, line 12: Better to prefix `AlarmEmailTo` with a payout or something as we might add more scripts in the future that might need to send email to different parties and will need a different prefix.

lovesh (approved):

SamHellawell:
@olegnn is this ready to be merged?

olegnn:
@SamHellawell yes, but please pay attention that there's also some small change in SDK code regarding initialization - [src/index.js](https://github.com/docknetwork/sdk/pull/286/files#diff-bfe9874d239014961b1ae4e89875a6155667db834a410aaaa2ebe3cf89820556).

Currently, if disconnect the `DockAPI` instance, you won't be able to initialize it again because `init` method assumes `anchorModule` to be initialized while on disconnect it will be deleted.
As far as the SDK itself exports `DockAPI` instance as `default` which assumes a singleton pattern, this use case may be very frequent.

Also, this function currently doesn't work as expected https://github.com/docknetwork/sdk/blob/master/src/index.js#L288 as it will return `anchorModule` after instantiation rather than after establishing a connection.

src/index.js
```diff
@@ -135,6 +134,7 @@ class DockAPI {
 
     await this.initKeyring(keyring);
 
+    this.anchorModule = new AnchorModule();
```
SamHellawell, line 137: instead of creating a new anchormodule instance each connect, would be better to simply remove this line https://github.com/docknetwork/sdk/blob/28f24976e53534a4477a3b627d67e9ffa6a3f8b8/src/index.js#L165 so that it isnt deleted.

good spot either way though

SamHellawell (commented):
looks good to me for most part, only minor thing is theres no need to recreate anchormodule instance (see comment). can approve/merge if we can address that quickly

src/index.js
olegnn: Won't be the problem that `anchorModule` will have disconnected `this.api` under the hood?
On `dock.anchor` getter call `AnchorModule` with disconnected API will be returned (https://github.com/docknetwork/sdk/blob/master/src/index.js#L288).

By the way, I'm not sure what's the problem with recreating `AnchorModule`? This class doesn't have a constructor and it doesn't extend any other class.

SamHellawell: the disconnected API will be returned yes, but there are parts of the anchor module that do not require an api connection to be able to use. thats why anchor module is constructed outside of init unlike the others (so its always available)

if you call `dockapi.init()` again after disconnect, it will call `setApi` in anchor module so that the latest api connection

SamHellawell (approved):

src/index.js
olegnn: I'd say it's better to either move these methods to some module without API dependency or make them static, but anyway I changed behavior to keep `anchorModule` always initiated.

SamHellawell: i agree, hopefully we can get some refactoring time here. thats quite a bit we had planned but SDK was put on into maintenance mode so we can focus on other projects

SamHellawell:
@olegnn would you mind, when theres time, to fix conflicts and merge?

SamHellawell:
thanks @olegnn !

PR #319, opened 2022-10-10
Did resolution returns bbs+ keys

lovesh:


src/modules/did/did.js
```diff
@@ -795,6 +796,47 @@ class DIDModule {
       }
     });
 
+    // Fetch BBS+ keys if needed
+    if (getBbsPlusSigKeys === true) {
```
SamHellawell, line 800: shouldnt need `getBbsPlusSigKeys === true` just `if (getBbsPlusSigKeys)`

src/modules/did/did.js
```diff
@@ -731,12 +732,12 @@ class DIDModule {
...
    */
   // eslint-disable-next-line sonarjs/cognitive-complexity
-  async getDocument(did) {
+  async getDocument(did, { getBbsPlusSigKeys = false } = {}) {
     const hexId = getHexIdentifierFromDID(did);
 
     let didDetails = await this.api.rpc.core_mods.didDetails(hexId, 15);
     if (didDetails.isNone) {
-      throw new NoDIDError(`dock:did:${did}`);
+      throw new NoDIDError(`dock:did:${hexDIDToQualified(hexId)}`);
```
SamHellawell, line 740: `dock:did:` should be `did:dock:` ?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@docknetwork/sdk",
-  "version": "2.4.2",
+  "version": "2.5.0",
```
SamHellawell, line 3: i feel like this change isnt breaking anything so 2.4.3 could work?

src/modules/did/did.js
```diff
@@ -795,6 +796,47 @@ class DIDModule {
       }
     });
 
+    // Fetch BBS+ keys if needed
```
SamHellawell, line 799: what would happen if:
- user has three did keys before this change (1, 2 and 3)
- user adds a BBS key

what would the index of the BBS key be? assuming 4, just want to make sure.

if user deletes a key after, would the bbs key still be index 4 or would it go to 3? if it goes to 3, there may be problems

SamHellawell (changes requested):

src/modules/did/did.js
lovesh: The argument `did` could be fully-qualified or just hex so can't rely on that.

package.json
lovesh: As per semver, patch version is for a non-breaking bugfix, non-breaking feature additions need a minor version change.

SamHellawell: fair enough, i guess i cheat a little and put some smaller patches into minor bumps. works for me

src/modules/did/did.js
lovesh: > assuming 4, just want to make sure.

Yes.

> if user deletes a key after, would the bbs key still be index 4

Remains 4. Key ids don't change and they behaving in the same way for BBS+ as for other types.

src/modules/did/did.js
SamHellawell: not sure what you mean, im referring to the error message

src/modules/did/did.js
lovesh: Wanted to be explicit

src/modules/did/did.js
SamHellawell: sounds good, just wanted to confirm

src/modules/did/did.js
SamHellawell: see the edit you made, yeah im aware after a quick search that dock:did is used - could still patch here too

lovesh: Done.

SamHellawell (approved):

PR #321, opened 2022-10-28
Hotfix for the upgraded DID RPC

olegnn:


src/modules/did/did.js
```diff
@@ -799,7 +799,8 @@ class DIDModule {
     // Fetch BBS+ keys if needed
     if (getBbsPlusSigKeys === true) {
       const details = didDetails.details.asOnChain;
-      const lastKeyId = details.lastKeyId.toNumber();
+      const data = details.data || details;
```
lovesh, line 802: We flatten `data` in `WithNonce` struct so `details.data` should not be needed,

olegnn: Autogenerated metadata doesn't support flattening, so this struct will have an explicit `data` field in the upgraded version.

lovesh (approved):

PR #325, opened 2022-11-08
Use credentialschema.essential if none defined in json

SamHellawell:


SamHellawell:
@lovesh have to merge this for the API, but if you'd like to discuss we can post-merge

PR #326, opened 2022-11-08
Fix BBS+ verify with schema

SamHellawell:
BBS+ credential verification was failing if schema resolution was enabled, have updated schema resolution to use the document loader which now supports embedded JSON

### DIFF
--- a/src/utils/vc/schema.js
+++ b/src/utils/vc/schema.js
@@ -28,7 +28,7 @@ export async function validateCredentialSchema(credential, schema, context, docu
       delete subject[credentialIDField];
     }
 
-    const compacted = await jsonld.compact(subject, context, { documentLoader }); // eslint-disable-line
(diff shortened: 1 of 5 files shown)


src/utils/vc/schema.js
```diff
@@ -28,7 +29,7 @@ export async function validateCredentialSchema(credential, schema, context) {
       delete subject[credentialIDField];
     }
 
-    const compacted = await jsonld.compact(subject, context); // eslint-disable-line
+    const compacted = await jsonld.compact(subject, context, { documentLoader: documentLoader || defaultDocumentLoader() }); // eslint-disable-line
```
lovesh, line 32: Why wasn't the `documentLoader` required before? This is just to resolve the context URLs, right?

lovesh (approved):

src/utils/vc/schema.js
SamHellawell: yeah to resolve context urls, not sure why it wasnt there before - guess an oversight thats been there a while

PR #329, opened 2022-11-11
Allow to verify credential expiration date

SamHellawell:


src/utils/vc/credentials.js
```diff
@@ -180,6 +181,27 @@ export async function verifyCredential(credential, {
   // Check credential is valid
   checkCredential(credential);
 
+  // Check expiration date
+  if (verifyDates && 'expirationDate' in credential) {
```
lovesh, line 185: It's better to move this check in `checkCredential`, probably in `checkCredentialOptional` where a valid expiration date is checked. By putting try-catch over `checkCredential`, following error message will cover other possible errors as well.

SamHellawell: my understanding was that checkCredential should just be checking the format, not the contents

lovesh (approved):

src/utils/vc/credentials.js
SamHellawell: > I am saying exactly that in my comment.

except if we do a try/catch around it for this invocation then it will be saying verified is false, instead of outright throwing an error here for passing an object that isnt a valid VC. the verifyCredential method should throw if its not valid content, but return verified true/false and proofs/errors as an object if its a valid VC and something is wrong, eg expired.

PR #332, opened 2022-12-07
BBS+ presentations

osinakayah:


src/bbs-plus-presentation.js
```diff
@@ -0,0 +1,103 @@
...
+   * @param j
+   * @param revealAttributes
+   * @returns {Promise<number>}
+   */
+  async addCredentialsToPresent(j, revealAttributes = []) {
+    ensureArray(revealAttributes);
+    await initializeWasm();
+    const json = typeof j === 'string' ? JSON.parse(j) : j;
+
+    const didDocument = await this.dock.did.getDocument(this.getCredentialIssuerDID(json));
```
SamHellawell, line 72: better to pass the publickey instead of quering the chain every time you add a credential to present, this class should theoretically work without blockchain interaction directly

src/bbs-plus-presentation.js
```diff
@@ -0,0 +1,103 @@
...
+
+    const pkRaw = b58.decode(didDocument.publicKey[1].publicKeyBase58);
+    const pk = new BBSPlusPublicKeyG2(pkRaw);
+
+    const { credentialSchema } = json;
+    const credSchema = credentialSchema ? CredentialSchema.fromJSON(
+      typeof credentialSchema === 'string'
+        ? JSON.parse(credentialSchema)
+        : credentialSchema,
+    ) : new CredentialSchema({
```
SamHellawell, line 82: shouldnt this be using schema generation? (lovesh's latest updates)

tests/integration/anoncreds/bbs-plus-presentation.test.js
```diff
@@ -0,0 +1,183 @@
...
+import { CredentialSchema, CredentialBuilder } from '@docknetwork/crypto-wasm-ts/lib/anonymous-credentials';
+import { DockAPI } from '../../../src';
+import { FullNodeEndpoint, TestAccountURI, TestKeyringOpts } from '../../test-constants';
+import { createNewDockDID } from '../../../src/utils/did';
+import Bls12381G2KeyPairDock2022 from '../../../src/utils/vc/crypto/Bls12381G2KeyPairDock2022';
+import BbsPlusPresentation from '../../../src/bbs-plus-presentation';
+import BBSPlusModule from '../../../src/modules/bbs-plus';
+import { registerNewDIDUsingPair } from '../helpers';
+
+const residentCardSchema1 = {
```
SamHellawell, line 13: better to move into example/test fixtures since its used in a few places

SamHellawell (changes requested):

src/bbs-plus-presentation.js
```diff
@@ -0,0 +1,103 @@
...
+    throw new Error('Unable to retrieve issuer DID');
+  }
+
+  /**
+   * Add jsonld credentials to be presented.
+   * @param j
+   * @param revealAttributes
+   * @returns {Promise<number>}
+   */
+  async addCredentialsToPresent(j, revealAttributes = []) {
```
SamHellawell, line 67: param `j` doesnt make much sense for naming

src/bbs-plus-presentation.js
SamHellawell: (should be now had to bump the crypto-wasm-ts version)

src/bbs-plus-presentation.js
```diff
@@ -0,0 +1,90 @@
...
+    const json = typeof j === 'string' ? JSON.parse(j) : j;
+
+    const didDocument = await this.dock.did.getDocument(this.getCredentialIssuerDID(json));
+
+    const pkRaw = b58.decode(didDocument.publicKey[1].publicKeyBase58);
+    const pk = new BBSPlusPublicKeyG2(pkRaw);
+
+    const { credentialSchema } = json;
+    const credSchema = credentialSchema ? CredentialSchema.fromJSON(
+      typeof credentialSchema === 'string'
```
SamHellawell, line 79: credentialSchema should never be a string. see https://github.com/docknetwork/sdk/blob/master/src/utils/vc/crypto/Bls12381BBSSignatureDock2022.js#L88

tests/integration/anoncreds/bbs-plus-presentation.test.js
```diff
@@ -0,0 +1,183 @@
...
+    await dock.disconnect();
+  }, 10000);
+});
+
+function getBBSCredential(did1, keypair) {
+  const credSchema = new CredentialSchema(residentCardSchema1, { useDefaults: true });
+  const credBuilder = new CredentialBuilder();
+  credBuilder.schema = credSchema;
+  credBuilder.subject = credentialJSONWithoutSchema.credentialSubject;
+  for (const k of ['@context', 'id', 'type', 'identifier', 'name', 'description']) {
```
SamHellawell, line 176: this wont take into account all fields

src/bbs-plus-presentation.js
SamHellawell: this is also incorrect as you are assuming the BBS+ key exists at `publicKey[1]` which isnt always the case

tests/integration/anoncreds/bbs-plus-presentation.test.js
```diff
@@ -0,0 +1,183 @@
...
+      bbsPlusPresentation.addAttributeToReveal(idx, {});
+    }).toThrow();
+  });
+  afterAll(async () => {
+    await dock.disconnect();
+  }, 10000);
+});
+
+function getBBSCredential(did1, keypair) {
+  const credSchema = new CredentialSchema(residentCardSchema1, { useDefaults: true });
```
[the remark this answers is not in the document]
SamHellawell: updated in this branch, you should issue credentials similar to the issuing test https://github.com/docknetwork/sdk/blob/master/tests/integration/anoncreds/issuing.test.js#L118 

otherwise presentation being valid here may be a false positive

src/bbs-plus-presentation.js
osinakayah: credentialSchema could be a stringified object right?

SamHellawell: not in a JSON-LD credential - the format is very slightly different between crypto-wasm-ts and w3c json ld credentials so requires some transforming to be able to call `Credential.fromJSON()`

you can see how it should be transformed https://github.com/docknetwork/sdk/blob/DCKA-1202-bbs-presentation-wrapper/src/utils/vc/crypto/Bls12381BBSSignatureDock2022.js#L92

src/utils/vc/crypto/Bls12381BBSSignatureDock2022.js
```diff
@@ -81,57 +82,71 @@ export default class Bls12381BBSSignatureDock2022 extends CustomLinkedDataSignat
...
+    return nameValues[1];
+  }
+
+  async serializeForSigning(options) {
+    return await Bls12381BBSSignatureDock2022.convertCredential(options);
   }
 
-  async serializeForSigning({
-    document, proof,
+  static async convertCredential({
```
lovesh, line 96: This doesn't need to be async.

src/utils/vc/crypto/Bls12381BBSSignatureDock2022.js
```diff
@@ -81,57 +82,71 @@ export default class Bls12381BBSSignatureDock2022 extends CustomLinkedDataSignat
...
+    const [serializedCredential, credSchema] = await this.serializeForSigning(options);
 
     // Encode messages, retrieve names/values array
-    const namesValues = credSchema.encoder.encodeMessageObject(serializedCredential, SIGNATURE_PARAMS_LABEL_BYTES);
-    return namesValues[1];
+    const nameValues = credSchema.encoder.encodeMessageObject(serializedCredential, SIGNATURE_PARAMS_LABEL_BYTES);
+    return nameValues[1];
+  }
+
+  async serializeForSigning(options) {
```
lovesh, line 92: Is this function needed as its calling another function with the same args without doing anything? 
It can be static and doesn't need to be async.

src/bbs-plus-presentation.js
```diff
@@ -0,0 +1,82 @@
...
+
+  /**
+   *
+   * @param presentationLD
+   * @param {Array.<string>} publicKeys
+   * @returns {Promise<boolean>}
+   */
+  async verifyPresentation(presentationLD, publicKeys) {
+    ensureArray(publicKeys);
+    await initializeWasm();
```
lovesh, line 69: I think we should make an assumption that `initializeWasm` has been called before calling anything relevant in the ts repo. This way your functions don't have to be async just because wasm needs to be intiialized. We already follow a similar approach when dealing with polkadot-js crypto utils.

src/utils/vc/crypto/Bls12381BBSSignatureDock2022.js
```diff
@@ -18,8 +18,9 @@ import CustomLinkedDataSignature from './custom-linkeddatasignature';
 
 const SUITE_CONTEXT_URL = 'https://www.w3.org/2018/credentials/v1';
 
-const DEFAULT_PARSING_OPTS = {
-  useDefaults: true,
+export const DEFAULT_PARSING_OPTS = {
+  // useDefaults: true,
```
lovesh, line 22: Why was this changed?

lovesh (changes requested):

src/utils/vc/crypto/Bls12381BBSSignatureDock2022.js
osinakayah: I think @SamHellawell changed it when he was debugging the issue.

src/utils/vc/crypto/Bls12381BBSSignatureDock2022.js
SamHellawell: it used to do more and was refactored into that shared static method, think we can remove it

src/utils/vc/crypto/Bls12381BBSSignatureDock2022.js
SamHellawell: was for debugging schemas at one point but I think it makes sense to have use defaults as false now because we have schema generation - so defaults are never used

src/bbs-plus-presentation.js
```diff
@@ -0,0 +1,81 @@
...
+    return this.presBuilder.addCredential(Credential.fromJSON(credential, CustomLinkedDataSignature.fromJsigProofValue(credentialLD.proof.proofValue)), pk);
+  }
+
+  /**
+   *
+   * @param presentationLD
+   * @param {Array.<string>} publicKeys
+   * @returns {boolean}
+   */
+  verifyPresentation(presentationLD, publicKeys) {
```
SamHellawell, line 67: this method assumes you know the public keys of the issuer(s) of the credential(s) - but if i were to receive this presentation JSON from the API i wont know who issued it. we need to include some property like `verificationMethod`. @lovesh any ideas?

src/bbs-plus-presentation.js
```diff
@@ -0,0 +1,81 @@
...
+  PresentationBuilder,
+  Credential,
+} from '@docknetwork/crypto-wasm-ts/lib/anonymous-credentials';
+import { Presentation } from '@docknetwork/crypto-wasm-ts/lib/anonymous-credentials/presentation';
+import { ensureArray } from './utils/type-helpers';
+
+import Bls12381BBSSignatureDock2022 from './utils/vc/crypto/Bls12381BBSSignatureDock2022';
+import CustomLinkedDataSignature from './utils/vc/crypto/custom-linkeddatasignature';
+
+export default class BbsPlusPresentation {
```
SamHellawell, line 13: this class doesnt let you provide a nonce

src/bbs-plus-presentation.js
lovesh: Include that in the `proof` field in the Credential like `type` and those are always revealed. We should make `proof` a special top level field like `credStatus` and the mandate the reveal of the fields like `type`, `verificationMethod`

src/bbs-plus-presentation.js
lovesh: And context as well.

src/bbs-plus-presentation.js
SamHellawell: we already have this field in proof for SDK issued credentials, so i guess in this wrapper we can just enforce to reveal it

lovesh: Yes.

SamHellawell, line 67: also @osinakayah this method shouldnt be in this class because it doesnt use the presentation builder, better to just export it as a `verifyPresentation` function. it should also be updated to accept a documentLoader parameter and get the issuer public keys from each credential's revealed `verificationMethod` field

SamHellawell: alternative you could modify this method https://github.com/docknetwork/sdk/blob/master/src/utils/vc/presentations.js#L104 to detect a BBS+ presentation and apply above feedback, this way we can use the same `verifyPresentation(...)` api. if not id personally put a new method here called `verifyBBSPresentation` just to keep things organised a bit better

osinakayah: What if the issuerDID or verificationMethod field isn't revealed, is there another way to get the issuer public keys?

SamHellawell: nope, thats why you must enforce its revealed in this wrapper code (eg when a credential is added, call revealField(proof.verificationMethod) which will give the key ID - pass that to documentLoader and it will load the key JSON object from which you can get the base58 encoded pubic key

src/bbs-plus-presentation.js
```diff
@@ -0,0 +1,69 @@
...
+
+  /**
+   * Create bbs presentation
+   * @param options
+   * @returns {object}
+   */
+  createPresentation(options = {}) {
+    const { nonce } = options;
+    if (nonce && typeof nonce === 'string') {
+      this.presBuilder.nonce = b58.decode(nonce);
```
SamHellawell, line 39: this doesnt make much sense to enforce nonce to always be b58 encoded, also doesnt account for if nonce is not string (what if user wants to pass uint8array directly?)

src/bbs-plus-presentation.js
```diff
@@ -0,0 +1,69 @@
...
+    ensureArray(attributes);
+    this.presBuilder.markAttributesRevealed(credentialIndex, new Set(attributes));
+  }
+
+  /**
+   * Create bbs presentation
+   * @param options
+   * @returns {object}
+   */
+  createPresentation(options = {}) {
```
SamHellawell, line 36: we should supply context here also

src/bbs-plus-presentation.js
SamHellawell: use the `stringToU8a` method from @polkadot/util (theres examples of this in the code if you search for it, quite simple method)

PR #335, opened 2022-12-16
Make compatible with `v39` mainnet, `v44` testnet

olegnn:


lovesh:
Shouldn't there be an updated fastblock image for the tests?

lovesh:
It should be possible to pass the image tag to the script `with_docker..` Then the CI  can run tests for each image tag in a separate job. This will speed up the CI and make it easier to spot which image had errors.
If it's a big effort, you can do it in a future PR.

package.json
```diff
@@ -78,15 +78,15 @@
     "test-eth": "NODE_ENV=production jest --verbose --runInBand tests/integration/eth",
     "docs": "mkdir out && touch out/.nojekyll && jsdoc src -r -c .jsdoc -d out/reference",
     "dev-node": "./scripts/run_node_in_docker --dev --rpc-external --ws-external --rpc-cors=all",
-    "test-with-node": "./scripts/with_docker_test_node yarn test-integration",
-    "test-eth-with-node": "./scripts/with_docker_test_node yarn test-eth",
-    "examples-with-node": "./scripts/with_docker_test_node yarn test-examples",
+    "test-with-node": "./scripts/with_all_docker_test_nodes yarn test-integration",
```
lovesh, line 81: Should be possible to run with a specific image only.

lovesh (approved):
Thanks for the change. The examples run with mainnet tag only but its less important as everything should be covered by tests. So good to merge now.

lovesh (approved):

PR #341, opened 2023-01-19
Fee tests

olegnn:


tests/integration/__snapshots__/fees.test.js.snap
```diff
@@ -0,0 +1,113 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
lovesh, line 1: File needs to be removed.

tests/integration/fees.test.js
```diff
@@ -0,0 +1,723 @@
...
+
+  test("democracy.notePreimage", async () => {
+    const runtimeVersion = await getRuntimeVersion();
+    const specVersion = runtimeVersion.specVersion.toNumber();
+
+    if (process.env.DISABLE_DEMOCRACY_PREIMAGE_TESTS || specVersion < 44) {
+      return;
+    }
+
+    const preimage = randomAsHex(2e6);
```
lovesh, line 656: Although you have explained me the intention of these tests in DM, it will be better to put a comment here about the same.

lovesh (approved):

example/didweb.js
```diff
@@ -1,3 +1,4 @@
+/*
```
lovesh, line 1: @SamHellawell The test fails, likely due to DID resolution.

tests/integration/fees.test.js
```diff
@@ -0,0 +1,723 @@
...
+
+describe("Fees", () => {
+  const dock = new DockAPI();
+
+  async function getBalance(account) {
+    const { data: balance } = await dock.api.query.system.account(account);
+    return balance;
+  }
+
+  function getDidPair() {
```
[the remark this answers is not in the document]
olegnn: It's used in scripts, and we don't share any helpers between scripts and tests. But I moved it to the tests `helpers`.

lovesh (approved):

PR #343, opened 2023-02-22
JWS issuance by default

SamHellawell:
BBS+ retains proofValue but other signing schemes return JWS by default in the proof. proofValue can be used if set with issue credential args


tests/unit/issuing.test.js
```diff
@@ -118,6 +118,13 @@ describe('Verifiable Credential Issuing', () => {
     expect(result.results[0].verified).toBe(true);
   }, 30000);
 
+  test('Issuing should return an object with a proof, and it must pass validation (using proofValue).', async () => {
```
mike-parkhill, line 121: do you need a test for the `.jws` value too?

lovesh (approved):

tests/unit/issuing.test.js
SamHellawell: the tests cover that already (see changes from proofValue -> jws)

lovesh (approved):

PR #345, opened 2023-03-24
JWT VCs and JsonWebKey2020 support

SamHellawell:
- Create JWT VC
- Verify JWT VC
- JsonWebKey2020 support (JWT and non-JWT)
- Updates issuance test to cover more key types
- Added test for pre-signed credentials for robustness

JWT VPs not supported in this PR to keep changes minimal

### DIFF
--- a/src/utils/vc/credentials.js
+++ b/src/utils/vc/credentials.js
@@ -46,10 +46,13 @@ export function formatToJWTPayload(keyDoc, cred) {
   const kid = keyDoc.id;
   const credentialIssuer = cred.issuer;
   const subject = cred.credentialSubject.id;
-  const validFrom = cred.validFrom || cred.issuanceDate;
   const { issuanceDate, expirationDate } = cred;
 
-  // Reference: https://www.w3.org/TR/vc-data-model/#jwt-encoding
+  // NOTE: Expecting validFrom here for future spec support
+  const validFrom = cred.validFrom || issuanceDate;
+
+  // References: https://www.w3.org/TR/vc-data-model/#jwt-encoding
+  // https://www.rfc-editor.org/rfc/rfc7519#section-4.1.6
   const vcJwtPayload = {
     jti: cred.id,
     sub: subject || '',
--- a/src/utils/vc/credentials.js
+++ b/src/utils/vc/credentials.js
@@ -1,5 +1,7 @@
 import jsonld from 'jsonld';
 import jsigs from 'jsonld-signatures';
+
+import base64url from 'base64url';
 import CredentialIssuancePurpose from './CredentialIssuancePurpose';
 import defaultDocumentLoader from './document-loader';
 import { getAndValidateSchemaIfPresent } from './schema';
@@ -12,8 +14,9 @@ import { ensureValidDatetime } from '../type-helpers';
 
 import {
   EcdsaSepc256k1Signature2019, Ed25519Signature2018, Sr25519Signature2020,
-  Bls12381BBSSignatureDock2022, Bls12381BBSSignatureProofDock2022,
+  Bls12381BBSSignatureDock2022, Bls12381BBSSignatureProofDock2022, JsonWebSignature2020,
 } from './custom_crypto';
+import { bufferToUint8Array, signJWS } from './crypto/custom-linkeddatasignature';
 
 /**
  * @param {string|object} obj - Object with ID property or a string
@@ -148,13 +151,13 @@ export function checkCredential(credential) {
 
 /**
  * Verify a Verifiable Credential. Returns the verification status and error in an object
- * @param {object} [credential] The VCDM Credential
+ * @param {object} [vcJSONorString] The VCDM Credential as JSON-LD or JWT string
  * @param {VerifiableParams} options Verify parameters, this object is passed down to jsonld-signatures calls
  * @return {Promise<object>} verification result. The returned object will have a key `verified` which is true if the
  * credential is valid and not revoked and false otherwise. The `error` will describe the error if any.
  */
-// TODO: perform verification on the expanded JSON-LD credential
-export async function verifyCredential(credential, {
+// eslint-disable-next-line sonarjs/cognitive-complexity
+export async function verifyCredential(vcJSONorString, {
   resolver = null,
   compactProof = true,
   forceRevocationCheck = true,
@@ -169,6 +172,9 @@ export async function verifyCredential(credential, {
   if (documentLoader && resolver) {
     throw new Error('Passing resolver and documentLoader results in resolver being ignored, please re-factor.');
   }
+
+  const isJWT = typeof vcJSONorString === 'string';
+  const credential = isJWT ? JSON.parse(base64url.decode(vcJSONorString.split('.')[1])).vc : vcJSONorString;
 
   if (!credential) {
     throw new TypeError(
@@ -213,13 +219,41 @@ export async function verifyCredential(credential, {
     await getAndValidateSchemaIfPresent(expandedCredential, schemaApi, credential[credentialContextField], docLoader);
   }
 
-  // Verify with jsonld-signatures
+  // JWT formatted credential?
+  if (isJWT) {
+    const header = JSON.parse(base64url.decode(vcJSONorString.split('.')[0]).toString());
+    if (!header.kid) {
+      throw new Error('No kid in JWT header');
+    }
+
+    const { document: keyDocument } = await docLoader(header.kid);
+    const keyDocSuite = await getSuiteFromKeyDoc(keyDocument, false, { detached: false, header });
+    const verified = await keyDocSuite.verifySignature({
+      verifyData: bufferToUint8Array(Buffer.from(vcJSONorString.split('.')[1], 'utf8')),
+      verificationMethod: keyDocument,
+      proof: {
+        jws: vcJSONorString,
+      },
+    });
+
+    return { verified };
+  }
+
+  // Verify with jsonld-signatures otherwise
   const result = await jsigs.verify(credential, {
     purpose: purpose || new CredentialIssuancePurpose({
       controller,
     }),
     // TODO: support more key types, see digitalbazaar github
-    suite: [new Ed25519Signature2018(), new EcdsaSepc256k1Signature2019(), new Sr25519Signature2020(), new Bls12381BBSSignatureDock2022(), new Bls12381BBSSignatureProofDock2022(), ...suite],
+    suite: [
+      new Ed25519Signature2018(),
+      new EcdsaSepc256k1Signature2019(),
+      new Sr25519Signature2020(),
+      new Bls12381BBSSignatureDock2022(),
+      new Bls12381BBSSignatureProofDock2022(),
+      new JsonWebSignature2020(),
+      ...suite,
+    ],
     documentLoader: docLoader,
     compactProof,
   });
@@ -250,17 +284,15 @@ export async function verifyCredential(credential, {
  *   `@context` (if it is not present, it's added to the context list).
  * @return {Promise<object>} The signed credential object.
  */
-export async function issueCredential(keyDoc, credential, compactProof = true, documentLoader = null, purpose = null, expansionMap = null, issuerObject = null, addSuiteContext = false, useProofValue = false) {
-  // Get suite from keyDoc parameter
-  const suite = getSuiteFromKeyDoc(keyDoc, useProofValue);
-  if (!suite.verificationMethod) {
-    throw new TypeError('"suite.verificationMethod" property is required.');
-  }
+export async function issueCredential(keyDoc, credential, compactProof = true, documentLoader = null, purpose = null, expansionMap = null, issuerObject = null, addSuiteContext = true, type = false) {
+  const useProofValue = type === 'proofValue';
+  const useJWT = type === 'jwt';
 
   // Clone the credential object to prevent mutation
   const issuerId = credential.issuer || keyDoc.controller;
   const cred = {
     ...credential,
+    // issued: credential.issuanceDate,
     issuer: issuerObject ? {
       ...issuerObject,
       id: issuerId,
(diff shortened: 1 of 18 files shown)


src/utils/vc/credentials.js
```diff
@@ -170,6 +173,9 @@ export async function verifyCredential(credential, {
     throw new Error('Passing resolver and documentLoader results in resolver being ignored, please re-factor.');
   }
 
+  const isJWT = typeof vcJSONorString === 'string';
+  const credential = isJWT ? JSON.parse(base64url.decode(vcJSONorString.split('.')[1])).vc : vcJSONorString;
```
mike-parkhill, line 177: are these array indices guaranteed? (same for the one on line 224)

src/utils/vc/credentials.js
```diff
@@ -250,17 +284,15 @@ export async function verifyCredential(credential, {
...
-  }
+export async function issueCredential(keyDoc, credential, compactProof = true, documentLoader = null, purpose = null, expansionMap = null, issuerObject = null, addSuiteContext = true, type = false) {
+  const useProofValue = type === 'proofValue';
+  const useJWT = type === 'jwt';
 
   // Clone the credential object to prevent mutation
   const issuerId = credential.issuer || keyDoc.controller;
   const cred = {
     ...credential,
+    // issued: credential.issuanceDate,
```
mike-parkhill, line 295: commented code?

src/utils/vc/credentials.js
```diff
@@ -250,17 +284,15 @@ export async function verifyCredential(credential, {
  *   `@context` (if it is not present, it's added to the context list).
  * @return {Promise<object>} The signed credential object.
  */
-export async function issueCredential(keyDoc, credential, compactProof = true, documentLoader = null, purpose = null, expansionMap = null, issuerObject = null, addSuiteContext = false, useProofValue = false) {
-  // Get suite from keyDoc parameter
-  const suite = getSuiteFromKeyDoc(keyDoc, useProofValue);
-  if (!suite.verificationMethod) {
-    throw new TypeError('"suite.verificationMethod" property is required.');
-  }
+export async function issueCredential(keyDoc, credential, compactProof = true, documentLoader = null, purpose = null, expansionMap = null, issuerObject = null, addSuiteContext = true, type = false) {
```
mike-parkhill, line 287: `type = false` feels weird to me. Was expecting a string, not a bool

src/utils/vc/credentials.js
```diff
@@ -270,7 +302,73 @@ export async function issueCredential(keyDoc, credential, compactProof = true, d
   // Ensure credential is valid
   checkCredential(cred);
 
-  // Sign and return the credential
+  // Should use JWT format?
+  if (useJWT) {
```
mike-parkhill, line 306: could we pull the useJWT code out into helper functions that we can write tests for more simply? I always get nervous with `if` statements in large blocks of code

src/utils/vc/crypto/custom-linkeddatasignature.js
```diff
@@ -4,20 +4,68 @@ import base64url from 'base64url';
 
 const MULTIBASE_BASE58BTC_HEADER = 'z';
 
-function createJws({ encodedHeader, verifyData }) {
+const detachedHeaderParams = {
+  b64: false,
+  crit: ['b64'],
+};
+
+// Taken from https://github.com/transmute-industries/verifiable-data/blob/main/packages/jose-ld/src/JWS/createSigner.ts
```
mike-parkhill, line 12: don't suppose they have a library we could just borrow instead of copying the code?

src/utils/vc/crypto/custom-linkeddatasignature.js
```diff
@@ -63,14 +111,11 @@ export default class CustomLinkedDataSignature extends jsigs.suites.LinkedDataSi
...
 
-      // confirm header matches all expectations
-      if (!(header.alg === this.alg && header.b64 === false
-        && Array.isArray(header.crit) && header.crit.length === 1
-        && header.crit[0] === 'b64')) {
-        throw new Error(
-          `Invalid JWS header parameters for ${this.type}.`,
-        );
-      }
+      // if (header.alg !== this.alg) {
```
mike-parkhill, line 114: commented code

tests/unit/issuing.test.js
```diff
@@ -33,471 +44,471 @@ function getSampleCredential(signed = false) {
...
+      const result = await verifyCredential(credential);
+      expect(result.verified).toBe(true);
+      expect(result.results[0].proof).toBeDefined();
+      expect(result.results[0].proof.proofPurpose).toBe('assertionMethod');
+      expect(result.results[0].proof.type).toBe(sigType);
+      expect(result.results[0].proof.verificationMethod).toBe(keyUrl);
+      expect(result.results[0].verified).toBe(true);
+    }, 30000);
+
+    // JsonWebKey signatures cant use proofValue
```
mike-parkhill, line 121: why not? sorry if it's a dumb question - I haven't looked into JWK's at all

SamHellawell: because the jsonwebkey signed only produces jws signatures, propfValue isn't one and can't be represented as one

src/utils/vc/credentials.js
SamHellawell: if it's a valid JWT format then yeah they are

src/utils/vc/credentials.js
SamHellawell: need to see if it makes sense to set this field in the SDK or if called should decide

src/utils/vc/credentials.js
SamHellawell: yeah this was just leftover from `proofValue = false` - will update

src/utils/vc/credentials.js
SamHellawell: its already split into helpers really:
```
    // Format to VC JWT spec
    const [vcJwtHeader, vcJwtPayload] = formatToJWTPayload(keyDoc, cred);

    // Get suite from keyDoc parameter
    const jwtOpts = { detached: false, header: vcJwtHeader };
    const suite = await getSuiteFromKeyDoc(keyDoc, useProofValue, jwtOpts);
    return signJWS(suite.signer || suite, suite.alg, jwtOpts, vcJwtPayload);
```

dont think this would be useful outside of this scope

src/utils/vc/crypto/custom-linkeddatasignature.js
SamHellawell: unfortunately they dont export just this method in their library

SamHellawell: theres also slight modifications

src/utils/vc/credentials.js
```diff
@@ -27,7 +30,39 @@ function getId(obj) {
...
+  const subject = cred.credentialSubject.id;
+  const validFrom = cred.validFrom || cred.issuanceDate;
+  const { issuanceDate, expirationDate } = cred;
+
+  // Reference: https://www.w3.org/TR/vc-data-model/#jwt-encoding
+  const vcJwtPayload = {
+    jti: cred.id,
+    sub: subject || '',
+    iss: credentialIssuer.id || credentialIssuer,
+    iat: Math.floor(Date.parse(issuanceDate) / 1000),
```
lovesh, line 48: The reference link above doesn't specify `iat`

src/utils/vc/credentials.js
```diff
@@ -27,7 +30,39 @@ function getId(obj) {
...
+    vc: cred,
+  };
+
+  if (validFrom) {
+    vcJwtPayload.nbf = Math.floor(Date.parse(validFrom) / 1000);
```
lovesh, line 53: The reference link above says to construct `nbf` from `issuanceDate` and not `validFrom`

src/utils/vc/credentials.js
```diff
@@ -27,7 +30,39 @@ function getId(obj) {
...
+  }
+
+  if (expirationDate) {
+    vcJwtPayload.exp = Math.floor(Date.parse(expirationDate) / 1000);
```
lovesh, line 57: You can abstract this to a function `Math.floor(Date.parse(expirationDate) / 1000)` and call that for this and above.

src/utils/vc/credentials.js
```diff
@@ -250,12 +321,9 @@ export async function verifyCredential(credential, {
  *   `@context` (if it is not present, it's added to the context list).
  * @return {Promise<object>} The signed credential object.
  */
-export async function issueCredential(keyDoc, credential, compactProof = true, documentLoader = null, purpose = null, expansionMap = null, issuerObject = null, addSuiteContext = false, useProofValue = false) {
-  // Get suite from keyDoc parameter
-  const suite = getSuiteFromKeyDoc(keyDoc, useProofValue);
-  if (!suite.verificationMethod) {
-    throw new TypeError('"suite.verificationMethod" property is required.');
-  }
+export async function issueCredential(keyDoc, credential, compactProof = true, documentLoader = null, purpose = null, expansionMap = null, issuerObject = null, addSuiteContext = true, type = 'jsonld') {
```
lovesh, line 324: What are the possible values of `type`? Its default value is `jsonld` but inside the function, it checks for only 2 values `proofValue` and `jwt` which suggests that the default value should be one of those

src/utils/vc/credentials.js
```diff
@@ -270,7 +338,24 @@ export async function issueCredential(keyDoc, credential, compactProof = true, d
...
 
-  // Sign and return the credential
+  // Should use JWT format?
+  if (useJWT) {
+    // Format to VC JWT spec
+    const [vcJwtHeader, vcJwtPayload] = formatToJWTPayload(keyDoc, cred);
+
+    // Get suite from keyDoc parameter
+    const jwtOpts = { detached: false, header: vcJwtHeader };
+    const suite = await getSuiteFromKeyDoc(keyDoc, useProofValue, jwtOpts);
```
lovesh, line 348: Dont need `useProofValue` as false can be passed.

src/utils/vc/crypto/custom-linkeddatasignature.js
```diff
@@ -4,20 +4,68 @@ import base64url from 'base64url';
...
 
-function decodeBase64Url(string) {
+export function decodeBase64Url(string) {
   const buffer = base64url.toBuffer(string);
   return new Uint8Array(buffer.buffer, buffer.offset, buffer.length);
 }
 
-function decodeBase64UrlToString(string) {
+export function bufferToUint8Array(buffer) {
+  return new Uint8Array(buffer.buffer, buffer.offset, buffer.length);
```
lovesh, line 65: `new Uint8Array(buffer)` should be fine here and in several above instances.

src/utils/vc/crypto/custom-linkeddatasignature.js
```diff
@@ -122,6 +161,12 @@ export default class CustomLinkedDataSignature extends jsigs.suites.LinkedDataSi
       const signatureBytes = await getSigBytes(verifyData);
       finalProof.proofValue = CustomLinkedDataSignature.encodeProofValue(signatureBytes);
     } else {
+      if (!this.alg) {
+        throw new Error('Suite doesn\'t contain required alg parameter');
+      }
+
+      // TODO: use signJWS
```
lovesh, line 168: Whats the reason for not calling `signJWS`  here but in `issueCredential`?

src/verifiable-credential.js
```diff
@@ -235,18 +235,19 @@ class VerifiableCredential {
...
+  async sign(keyDoc, compactProof = true, issuerObject = null, addSuiteContext = true, format = null) {
     const signedVC = await issueCredential(
       keyDoc,
       this.toJSON(),
       compactProof,
       null, null, null,
       issuerObject,
       addSuiteContext,
-      useProofValue,
+      format,
```
lovesh, line 246: Nitpick: `issueCredential` calls it `type`

src/utils/vc/crypto/custom-linkeddatasignature.js
SamHellawell: agree, we had just used that other constructor before. will update

src/utils/vc/credentials.js
```diff
@@ -270,7 +338,24 @@ export async function issueCredential(keyDoc, credential, compactProof = true, d
...
+    return signJWS(suite.signer || suite, suite.alg, jwtOpts, vcJwtPayload);
+  }
+
+  // Get suite from keyDoc parameter
+  const suite = await getSuiteFromKeyDoc(keyDoc, useProofValue);
```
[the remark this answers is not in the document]
SamHellawell: if passing true here instead of `useProofValue` then all suites will sign with `proofValue` - which isnt intended

src/utils/vc/credentials.js
SamHellawell: well, logically it suggests that jsonld is the default type and others are checked incase. it can be all 3 of them - ill make them into constants and add some documentation strings

src/utils/vc/credentials.js
```diff
@@ -250,12 +321,9 @@ export async function verifyCredential(credential, {
...
  */
-export async function issueCredential(keyDoc, credential, compactProof = true, documentLoader = null, purpose = null, expansionMap = null, issuerObject = null, addSuiteContext = false, useProofValue = false) {
-  // Get suite from keyDoc parameter
-  const suite = getSuiteFromKeyDoc(keyDoc, useProofValue);
-  if (!suite.verificationMethod) {
-    throw new TypeError('"suite.verificationMethod" property is required.');
-  }
+export async function issueCredential(keyDoc, credential, compactProof = true, documentLoader = null, purpose = null, expansionMap = null, issuerObject = null, addSuiteContext = true, type = 'jsonld') {
+  const useProofValue = type === 'proofValue';
+  const useJWT = type === 'jwt';
```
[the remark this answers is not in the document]
SamHellawell: this isnt true because the `jsonld` format that doesnt use `proofValue` is default

src/utils/vc/crypto/custom-linkeddatasignature.js
SamHellawell: its an old todo actually, isnt needed here anymore (as signing is done by other crypto methods). `signJWS` is called in `issueCredential` because JWT VCs dont go through the linked data signature suite

src/utils/vc/credentials.js
SamHellawell: `iat` is part of general JWTs, ill add another link

src/utils/vc/credentials.js
SamHellawell: `validFrom` is `issuanceDate` if not supplied - please read VC data model disclaimer:
```
It is expected that the next version of this specification will add the validFrom [property](https://www.w3.org/TR/vc-data-model/#dfn-property) and will deprecate the issuanceDate [property](https://www.w3.org/TR/vc-data-model/#dfn-property) in favor of a new issued [property](https://www.w3.org/TR/vc-data-model/#dfn-property). The range of values for both properties are expected to remain as [[XMLSCHEMA11-2](https://www.w3.org/TR/xmlschema11-2/#dateTime)] combined date-time strings. Implementers are advised that the validFrom and issued [properties](https://www.w3.org/TR/vc-data-model/#dfn-property) are reserved and use for any other purpose is discouraged.
```

just trying to future proof

src/utils/vc/crypto/custom-linkeddatasignature.js
SamHellawell: ive also moved signJWS/createJWS into their own file as they ideally shouldnt be here

SamHellawell:
@lovesh addressed and replied to feedback, sorry took so long only just was able to get around to it

src/utils/vc/credentials.js
lovesh: Can you add this note in the code?

src/utils/vc/credentials.js
lovesh: Thanks. An enum-like type for `type` will help

src/utils/vc/credentials.js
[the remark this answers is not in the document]
lovesh: I see. Was assuming that there are only 2 possible values.

lovesh:
Thanks

src/utils/vc/credentials.js
SamHellawell: this was updated

src/utils/vc/credentials.js
SamHellawell: updated

lovesh (approved):

PR #365, opened 2023-09-27
Assign anoncreds schema to issued credentials

cykoder:


src/utils/vc/crypto/common/DockCryptoSignature.js
```diff
@@ -135,7 +133,18 @@ export default withExtendedStaticProperties(
         // }
       }
 
+      // Else, schema object exists but no ID means the SDK is signalling for the suite to generate a schema
+      if (!credSchema && document.credentialSchema) {
+        credSchema = new CredentialSchema(CredentialSchema.essential());
+      }
+
+      // Else, no schema was found so just use the essentials and v0.0.1 schema version
+      // NOTE: version is important here and MUST be 0.0.1 otherwise it will invalidate BBS+ credentials
```
mike-parkhill, line 142: can we have a test that enforces the schema version?

could there ever be a need to have different version? If so, what does that look like?

cykoder: we already have a test to make sure verifying that works, we dont support issuing 0.0.1/no defined schema after this PR

src/utils/vc/crypto/common/DockCryptoSignature.js
```diff
@@ -113,15 +114,12 @@ export default withExtendedStaticProperties(
...
+      if (document.credentialSchema && document.credentialSchema.id) {
         credSchema = CredentialSchema.fromJSON({
           // Passing all the default parsing options. Ideally `document.credentialSchema` should contain these
-          parsingOptions: {
-            useDefaults: false,
-            defaultMinimumInteger: -((2 ** 32) - 1),
-            defaultMinimumDate: -((2 ** 44) - 1),
-            defaultDecimalPlaces: 0,
-          },
+          parsingOptions: DefaultSchemaParsingOpts,
```
lovesh, line 122: If `document.credentialSchema.id` is present, it means the schema is already present so the schema's parsing optons should be used and not default because they could be different.

cykoder: if they are present they will be used, if not they will use defaults...

cykoder: i think you are missing the point that at this point in code it could be a json-ld document with no parsing opts set, and they will be set after the signing

lovesh: OK

lovesh (approved):

PR #370, opened 2023-10-06
Remove references to forceRevocationCheck

cykoder:


src/utils/vc/presentations.js
```diff
@@ -106,8 +106,6 @@ export async function verifyPresentationCredentials(
  * @property {DIDResolver} [resolver] - Resolver to resolve the issuer DID (optional)
  * @property {Boolean} [unsignedPresentation] - Whether to verify the proof or not
  * @property {Boolean} [compactProof] - Whether to compact the JSON-LD or not.
- * @property {Boolean} [forceRevocationCheck] - Whether to force revocation check or not.
```
lovesh, line 109: We removed this in favour of `skipRevocationCheck`. Another missing option is `skipSchemaCheck`. See [here](https://github.com/docknetwork/sdk/blob/master/src/utils/vc/credentials.js#L223) for details. Can you please add those docs here? Thanks

lovesh (approved):

src/utils/vc/presentations.js
cykoder: will spin off an issue and make a separate PR when have a bit of spare time

PR #373, opened 2023-10-20
Schema generation fixes for credentials and presentations, support fo…

lovesh:
…r schema URI and private status list cred


lovesh:
Some of these test failures are due to URLs like `https://w3id.org/vc/status-list/2021/v1` not being resolved.

cykoder (approved):
lgtm, might be good to get @olegnn's review also as it touches some code he produces

src/utils/vc/credentials.js
```diff
@@ -466,22 +472,200 @@ export async function issueCredential(
...
+    if (!verifyResult.verified) {
+      throw new Error(`Status list credential failed to verify with error: ${verifyResult.error}`);
+    }
+  }
+
+  // decode list from SL VC
+  // get JSON StatusList
+  const { credentialSubject: sl } = privateStatusListCredential;
+  const { encodedList } = sl;
+  const list = await decodeList({ encodedList });
```
olegnn, line 580: Wouldn't be better to accept `privateStatusListCredential` as an instance of `PrivateStatusList2021Credential` and use `await privateStatusListCredential.decodedStatusList()` instead?

olegnn (approved):
Great

src/utils/vc/credentials.js
lovesh: Yes. I had missed the caching aspect. Fixed now. Thanks

cykoder (approved):

PR #375, opened 2023-10-31
Ignore accumulator status type check in verifyCredential

lovesh:


gasher (dismissed):

src/utils/vc/credentials.js
```diff
@@ -397,7 +397,11 @@ export async function verifyCredential(
         }
       }
 
-      if (!isStatusList2021Status && !isRegRevStatus && getPrivateStatus(credential) === undefined) {
+      // For credentials supporting revocation with accumulator, the revocation check happens using witness which is not
+      // part of the credential and evolves over time
+      const isAccumStatus = isAccumulatorRevocationStatus(status);
+
+      if (!isStatusList2021Status && !isRegRevStatus && getPrivateStatus(credential) === undefined && isAccumStatus === false) {
```
olegnn, line 404: It would be better to have all `is...Status` as booleans and then check for 
```javascript
!isPrivateStatus && !isAccumStatus
```
instead of
```javascript
getPrivateStatus(credential) === undefined && isAccumStatus === false
```

olegnn (commented):
Otherwise LGTM

src/utils/vc/credentials.js
lovesh: Thanks. Done.

olegnn (approved):

cykoder (approved):

PR #382, opened 2023-12-12
`Trust registry` and `did:key`

olegnn:


src/utils/did.js
```diff
@@ -3,15 +3,128 @@
...
+import { Signature } from "../signatures"; // eslint-disable-line
+import {
+  getPublicKeyFromKeyringPair,
+  getSignatureFromKeyringPair,
+  getStateChange,
+} from './misc';
 
 export const DockDIDMethod = 'dock';
 export const DockDIDQualifier = `did:${DockDIDMethod}:`;
+export const DockDIDMethodKeyQualifier = `did:key:${DockDIDMethod}:`;
```
lovesh, line 19: The full DID would be `did:key:dock:<value>` but as per spec https://w3c-ccg.github.io/did-method-key/, full DID should be `did:key:<value>`

src/utils/did.js
```diff
@@ -3,15 +3,128 @@
...
+  publicKey() {
+    return getPublicKeyFromKeyringPair(this.keyPair);
+  }
+
+  sign(message) {
+    return getSignatureFromKeyringPair(this.keyPair, message);
+  }
+}
+
+const SECP256K1_PUBLIC_KEY_PREFIX = 'zQ3';
```
lovesh, line 39: `ZQ3` -> `ZQ3s` as per [spec](https://w3c-ccg.github.io/did-method-key/#secp256k1).

src/utils/did.js
```diff
@@ -3,15 +3,130 @@
...
+  }
+
+  signStateChange(name, payload, keyRef) {
+    const stateChange = getStateChange(this.api, name, payload);
+    const keySignature = keyRef.sign(stateChange);
+
+    return createDidSig(this, keyRef, keySignature);
+  }
+
+  changeState(method, name, payload, keyRef) {
```
lovesh, line 73: This is only used to create an extrinsic so better to name it accordingly.

src/utils/did.js
```diff
@@ -3,15 +3,130 @@
...
+  sign(message) {
+    return getSignatureFromKeyringPair(this.keyPair, message);
+  }
+}
+
+const SECP256K1_PUBLIC_KEY_PREFIX = 'zQ3';
+const ED_25519_PUBLIC_KEY_PREFIX = 'z6Mk';
+
+const DockDidMethodKeySecp256k1Prefix = `${DockDIDQualifier}${SECP256K1_PUBLIC_KEY_PREFIX}`;
+const DockDidMethodKeyEd25519Prefix = `${DockDIDMethodKeyQualifier}${ED_25519_PUBLIC_KEY_PREFIX}`;
```
lovesh, line 43: This and above should have `key` prefix and not `dock`

src/modules/trust-registry.js
```diff
@@ -0,0 +1,124 @@
...
+      convenerOrIssuerOrVerifierHexDid.changeState(
+        this.module.updateSchemaMetadata,
+        'UpdateSchemaMetadata',
+        { registryId, schemas, nonce: lastNonce },
+        signingKeyRef,
+      ),
+      waitForFinalization,
+      params,
+    );
+  }
```
lovesh, line 123: What about delegated issuers and suspension of issuers?

tests/integration/trust-registry-module.test.js
```diff
@@ -0,0 +1,326 @@
+import { randomAsHex } from "@polkadot/util-crypto";
```
lovesh, line 1: This test mentions status list repeatedly in the comments. Need to be fixed.

tests/integration/status-list-credential-module.test.js
```diff
@@ -91,149 +102,152 @@ buildTest('StatusListCredential Module', () => {
...
+        fetchedCred,
+        ownerDID,
+        pair,
+        { didModule: dock.did }
+      );
     await dock.statusListCredential.updateStatusListCredential(
       update,
-      [[sig, nonce]],
-      false,
+      [[nonce, sig]],
```
lovesh, line 212: Why change the order and introduce another breaking change? Better to keep the serialization of DIDSig the same on the node.

tests/integration/revocation.test.js
```diff
@@ -69,15 +70,15 @@ describe('Revocation Module', () => {
   }, 40000);
 
   test('Can revoke single from a registry', async () => {
-    const [revoke, sig, nonce] = await dock.revocation.createSignedRevoke(registryId, revokeIds, ownerDID, pair, 1, { didModule: dock.did });
-    await dock.revocation.revoke(revoke, [[sig, nonce]], false);
+    const [revoke, sig, nonce] = await dock.revocation.createSignedRevoke(registryId, revokeIds, ownerDID, pair, { didModule: dock.did });
+    await dock.revocation.revoke(revoke, [[nonce, sig]], false);
```
lovesh, line 74: The order should be kept same as before

src/modules/did/did.js
```diff
@@ -49,7 +51,7 @@ class DIDModule {
    * @returns {*}
    */
   createNewOffchainTx(did, didDocRef) {
-    const hexId = getHexIdentifierFromDID(did);
+    const hexId = typedHexDID(this.api, did).asDid;
```
lovesh, line 54: You know this did won't be a `did:key` so its more efficient to use `getHexIdentifier(did, DockDIDQualifier, DockDIDByteSize)` here than calling `typedHexDID` which creates an unnecessary object

src/modules/did/did.js
```diff
@@ -76,7 +78,7 @@ class DIDModule {
    * @returns {*}
    */
   createSetOffchainDidRefTx(did, didDocRef) {
-    const hexId = getHexIdentifierFromDID(did);
+    const hexId = typedHexDID(this.api, did).asDid;
```
lovesh, line 81: Similar comment as above on this line and similar lines as below.

src/modules/WithParamsAndPublicKeys.js
```diff
@@ -258,20 +252,20 @@ export default class WithParamsAndPublicKeys {
...
    * @param did
-   * @param keyId
+   * @param signingKeyRef
    * @param withParams - If true, return the params referenced by the public key. It will throw an error if paramsRef is null
    * or params were not found on chain which can happen if they were deleted after this public key was added.
    * @returns {Promise<{bytes: string}|null>}
    */
   async getPublicKey(did, keyId, withParams = false) {
-    const hexId = getHexIdentifierFromDID(did);
+    const hexId = typedHexDID(this.api, did);
```
lovesh, line 268: `did:key` can't have an onchain public key so you can just use `getHexIdentifier` and avoid creating extra object in `typedHexDID`

src/modules/schema.js
```diff
@@ -3,7 +3,6 @@ import { validate } from 'jsonschema';
 import axios from 'axios';
```
lovesh, line 3: axios was removed [here](https://github.com/docknetwork/sdk/pull/378).

src/utils/vc/credentials.js
```diff
@@ -3,8 +3,7 @@ import jsigs from 'jsonld-signatures';
 import { statusTypeMatches, checkStatus } from '@digitalbazaar/vc-status-list';
```
lovesh, line 3: This file is missing some code in `verifyCredential` from master branch.

src/modules/did/did.js
```diff
@@ -730,29 +682,35 @@ class DIDModule {
    * Throws NoDID if the DID does not exist on chain.
    * @param {string} did - The DID can be passed as fully qualified DID like `did:dock:<SS58 string>` or
    * a 32 byte hex string
+   * @param getOffchainSigKeys
    * @return {Promise<object>} The DID document.
    */
   // eslint-disable-next-line sonarjs/cognitive-complexity
   async getDocument(did, { getOffchainSigKeys = true } = {}) {
```
lovesh, line 689: Can you please update this method to use `didDetails` RPC so avoid multiple calls to the chain?

src/modules/did/did.js
olegnn: I don't think it will differ in efficiency despite an unnecessary object will be created.

src/modules/did/did.js
olegnn: This should be done separately.

src/utils/did.js
```diff
@@ -76,23 +196,64 @@ export function validateDockDIDSS58Identifier(identifier) {
...
+  const hex = getHexIdentifier(
+    u8aToHex(did.isDid ? did.asDid : did.asDidMethodKey),
+    [],
+    DockDIDByteSize,
+  );
+
+  if (did.isDid) {
+    return new DockDid(hex);
+  } else if (did.isDidMethodKey) {
+    return new DockDidMethodKey(hex);
```
lovesh, line 238: In `typedHexDID`, a `PublicKey` object is passed to `DockDidMethodKey` but here hex string is passed, why is that? What is the expectation from this function's returned value?

src/utils/did.js
```diff
@@ -76,23 +196,64 @@ export function validateDockDIDSS58Identifier(identifier) {
...
 /**
  * Return a fully qualified Dock DID id, i.e. "did:dock:<SS58 string>"
- * @param {string} hexId - The hex blob id (without the qualifier)
+ * @param {string|object} hexDid - The hex blob id (without the qualifier)
  * @returns {string} - The fully qualified Blob id
  */
-export function hexDIDToQualified(hexId) {
-  const ss58Id = encodeAddress(hexId);
-  return `${DockDIDQualifier}${ss58Id}`;
+export function hexDIDToQualified(hexDid) {
```
lovesh, line 249: Should be renamed as per the changed logic.

src/modules/trust-registry.js
```diff
@@ -0,0 +1,264 @@
...
+   * @param govFramework
+   * @param signingKeyRef
+   * @param nonce
+   * @param didModule
+   * @param params
+   * @param waitForFinalization
+   * @param params
+   * @returns {Promise<null>}
+   */
+  async init(
```
lovesh, line 35: This and the remaining methods in this call convert the actor DID to a typed one and get the nonce so the following lines are repeated in all methods
```
... = typedHexDID(..)
const lastNonce = nonce ?? (await getDidNonce(....
```

Better to move this to a separate function.

src/utils/did.js
```diff
@@ -76,23 +245,127 @@ export function validateDockDIDSS58Identifier(identifier) {
...
+    return getHexIdentifier(did, DockDIDQualifier, DockDIDByteSize);
+  } else if (did.isDid) {
+    const hex = getHexIdentifier(u8aToHex(did.asDid), [], DockDIDByteSize);
+
+    return new DockDid(hex);
+  } else if (did.isDidMethodKey) {
+    const key = did.asDidMethodKey;
+
+    if (key.isSecp256k1) {
+      const hex = getHexIdentifier(
```
lovesh, line 333: As per the implementation of `DockDidMethodKey`, the following 2 calls to it should fail. Is there a test for this branch of code?

lovesh (dismissed):

cykoder (approved):

mike-parkhill (approved):

PR #388, opened 2024-02-05
Tweaks for `did:key:*` and `did:dock:*` 

olegnn:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@docknetwork/sdk",
-  "version": "7.2.1",
+  "version": "7.2.2",
```
cykoder, line 3: this PR contains breaking changes, eg:
`DockDidMethodKey` -> `DidMethodKey`

so the version should be bumped up higher

PR #394, opened 2024-02-14
fix/bbs+ revocation with deriveCredential

maycon-mello:


cykoder:
@maycon-mello i have fixed your test (it was failing due to a few reasons, namely incorrect input data) and also that the credentialStatus wasn't persisting. I'm not sure if this alone is enough to ensure the full verification flow works, would like to see @lovesh's input there - I think we will need another E2E test for the full flow with proper accumulator setup and verification after derivation.

cykoder (approved):

cykoder (dismissed):

cykoder (approved):

PR #396, opened 2024-02-19
Align with the latest Trust Registry implementation

olegnn:


tests/integration/trust-registry-module.test.js
```diff
@@ -1,27 +1,27 @@
-import { randomAsHex } from '@polkadot/util-crypto';
```
cykoder, line 1: not sure why linting settings are off here, but all these could be single quotes not double to align with everywhere else

cykoder (approved):
one minor test linting comment but looks good otherwise

cykoder (approved):

PR #414, opened 2024-04-09
Add chain support for KB universal accum. and keyless accum

lovesh:


tests/integration/resolvers.test.js
```diff
@@ -61,21 +60,18 @@ describe('Resolvers', () => {
...
     await registerNewDIDUsingPair(dock, ownerDID, pair);
 
-    if (!DisableStatusListTests) {
-      statusListCred = await StatusList2021Credential.create(
-        ownerKey,
+    statusListCred = await StatusList2021Credential.create(
+      ownerKey,
+      statusListCredId,
+    {},
+    );await expect(
```
olegnn, line 67: `yarn lint-tests --fix` might be helpful.

src/utils/revocation.js
```diff
@@ -85,7 +85,12 @@ export const isRegistryRevocationStatus = ({ [credentialTypeField]: type }) => t
  * @param status
  * @returns {boolean}
  */
-export const isAccumulatorRevocationStatus = ({ [credentialTypeField]: type }) => type.includes(VB_ACCUMULATOR_22) || type.includes(`${LD_SEC_TERM}${VB_ACCUMULATOR_22}`) || type.includes(`/${VB_ACCUMULATOR_22}`);
+export const isAccumulatorRevocationStatus = ({ [credentialTypeField]: type }) => type.includes(VB_ACCUMULATOR_22)
```
olegnn, line 88: We could rewrite this code as

```js
const ACCUMULATOR_TYPES = [VB_ACCUMULATOR_22, `${LD_SEC_TERM}${VB_ACCUMULATOR_22}`, ...];

export const isAccumulatorRevocationStatus = ({ [credentialTypeField]: type }) => ACCUMULATOR_TYPES.some(accType => type.includes(accType));
```

src/modules/accumulator.js
```diff
@@ -79,6 +99,46 @@ export default class AccumulatorModule extends WithParamsAndPublicKeys {
     return null;
   }
 
+  /**
+   * Return the accumulated value as hex
+   * @param accumulated {Uint8Array|KBUniversalAccumulatorValue}
+   * @param typ {number} - Type of the accumulator
+   * @returns {string}
+   */
+  static accumulatedAsHex(accumulated, typ = AccumulatorType.VBPos) {
```
olegnn, line 108: I would be better to define all accumulated variants as classes like it's done with `KBUniversalAccumulatorValue `.
They can extend a class that has `fromHex`/`toHex` methods and encapsulate all this logic.

tests/test-constants.js
```diff
@@ -78,12 +78,12 @@ export const TestAccountCouncilMemberURI = fromEnv(
 export const MinGasPrice = fromEnv('MinGasPrice', DefaultMinGasPrice);
 export const MaxGas = fromEnv('MaxGas', DefaultMaxGas);
 export const TestSchemes = fromEnv('TestSchemes', DefaultTestSchemes);
-export const DisableStatusListTests = boolEnv(
-  fromEnv('DisableStatusListTests', 'false'),
-);
 export const DisableDidKeyAndTrustRegistryTests = boolEnv(
   fromEnv('DisableDidKeyAndTrustRegistryTests', 'false'),
 );
+export const DisableNewAccumulatorTests = boolEnv(
```
olegnn, line 84: Maybe it's better to call this variable `DisableKBAccumulatorTests`?

olegnn (changes requested):

src/modules/accumulator.js
lovesh: Thanks. Done

src/utils/revocation.js
lovesh: Thanks. Done

src/modules/accumulator.js
lovesh: It would be better to make this change in the TS dependency so i'll leave it like this here.

tests/test-constants.js
lovesh: Its for the keyless VB accum. as well.

olegnn (approved):

PR #418, opened 2024-04-22
Allow to provide node address as array for fallbacks

cykoder:


src/dock-api.js
```diff
@@ -99,10 +104,11 @@ export default class DockAPI {
...
     }
 
-    const isWebsocket = this.address && this.address.indexOf('http') === -1;
+    const addressStr = addressArray[0];
+    const isWebsocket = addressStr && addressStr.indexOf('http') === -1;
     const provider = isWebsocket
-      ? new WsProvider(this.address)
-      : new HttpProvider(this.address);
+      ? new WsProvider(addressArray)
+      : new HttpProvider(addressArray);
```
lovesh, line 111: `HttpProvider` doesn't accept an array as per its docs. Have you checked that it works?

cykoder: just updated this, thought maybe we had a test for it (we should at some point) but doesnt look like it

cykoder: created https://github.com/docknetwork/sdk/issues/419 for testing it in the future

src/dock-api.js
```diff
@@ -99,10 +104,16 @@ export default class DockAPI {
       rpc = Object.assign(rpc, PoaRpcDefs);
     }
 
-    const isWebsocket = this.address && this.address.indexOf('http') === -1;
+    const addressStr = addressArray[0];
```
lovesh, line 107: The correct way to handle would be to raise error if a mix of URL types is provided or accept a preference of websocket vs http. This requires a non-trivial amount of work so approving it for now. Can you add a note here please?

cykoder: yeah noticed that too, i was thinking of looking into a way to mix websocket/https urls if possible but im not sure polkadotjs allows this. essentially creating a provider per type/url

lovesh (approved):

PR #421, opened 2024-04-23
Skip schema check for anoncreds derived credentials

cykoder:
Skip JSON schema validation when verifying derived credentials, as anoncreds will do its own schema checks. This resolves an issue whereby schema validation fails for required fields that are not revealed.

### DIFF
--- a/src/utils/vc/credentials.js
+++ b/src/utils/vc/credentials.js
@@ -39,9 +39,12 @@ import {
   Bls12381BBSSignatureProofDock2022,
   Bls12381BBSSignatureDock2023,
   Bls12381BBSSignatureProofDock2023,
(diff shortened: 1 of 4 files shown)


src/utils/vc/credentials.js
```diff
@@ -293,7 +309,8 @@ export async function verifyCredential(
     documentLoader: docLoader,
   });
 
-  if (!skipSchemaCheck) {
+  const isAnoncredsDerived = isAnoncredsProofType(credential);
+  if (!skipSchemaCheck && !isAnoncredsDerived) {
```
lovesh, line 313: A comment explaining here the reason why `isAnoncredsDerived` is used to skip the check would be helpful.

And can you add a test for this?

lovesh (approved):

PR #430, opened 2024-06-10
Add accumulator update helpers and add test

lovesh:


src/modules/accumulator.js
```diff
@@ -927,6 +927,49 @@ export default class AccumulatorModule extends WithParamsAndPublicKeys {
...
+  // eslint-disable-next-line sonarjs/cognitive-complexity
+  async updateVbAccumulatorWitnessFromUpdatesInBlocks(accumulatorId, member, witness, startBlock, endBlock = undefined) {
+    if (endBlock === undefined) {
+      const accum = await this.accumulatorModule.getAccumulator(accumulatorId, false);
+      // eslint-disable-next-line no-param-reassign
+      endBlock = accum.lastModified;
+    }
+    console.debug(`Will start updating witness from block ${startBlock} to ${endBlock}`);
+    let current = startBlock;
+    while (current <= endBlock) {
```
cykoder, line 948: it would be faster to step this in multiple increments rather than one at a time, eg:
- step from current to +5 blocks
- await Promise.all results from those 5 blocks, process results
- continue to step next 5 or less blocks until end

currently its waiting on promise results for one block at a time rather than "chunks" of block results

cykoder (approved):

lovesh:
@cykoder A test fails due to change made in trust registries in the master branch of substrate node. This will be fixed once the corresponding change is made in SDK as well. Is this PR fine to merge since no active deployment is using the master branch (of node)?

PR #434, opened 2024-06-26
Refactor `typed-did`

olegnn:


cykoder (dismissed):
massive PR so may have missed some things but i think lgtm, however probably good to get @lovesh to look also

cykoder (approved):

PR #435, opened 2024-06-29
Change schema format from data-uri to object

lovesh:


cykoder:
Are there any tests that the current credentials with embedded data uri schema still validate and work? Such as https://github.com/docknetwork/sdk/blob/master/tests/data/static-bbs-cred-610.json (these dont have a schema set)

example/bbs-dock.js
```diff
@@ -63,8 +65,9 @@ const residentCardSchema = {
 };
 
 const embeddedSchema = {
-  id: `data:application/json;charset=utf-8,${encodeURIComponent(JSON.stringify(residentCardSchema))}`,
+  id,
   type: 'JsonSchemaValidator2018',
+  details: stringify({ jsonSchema: residentCardSchema }),
```
cykoder, line 70: if the ID references the schema now (i.e can be resolved) - why do we still need to stringify it here?

lovesh: This example is showing an embedded schema. I have set the `id` to the embedded schema's id so that schema could be identified by PEX filter. [Here](https://github.com/docknetwork/sdk/pull/435/files#diff-255c77598b22bdb3552345ef44b6f544957abf45ea0bc32027224391f83594f4R91) is the test using external schema.

lovesh:
> Are there any tests that the current credentials with embedded data uri schema still validate and work? Such as https://github.com/docknetwork/sdk/blob/master/tests/data/static-bbs-cred-610.json (these dont have a schema set)

[These](https://github.com/docknetwork/sdk/blob/schema-format/tests/integration/anoncreds/bbs-wallet.test.js) check credentials (0.5.0, 0.4.0 and an even older version) issued on testnet.

cykoder:
> [These](https://github.com/docknetwork/sdk/blob/schema-format/tests/integration/anoncreds/bbs-wallet.test.js) check credentials (0.5.0, 0.4.0 and an even older version) issued on testnet.

Would be good to move into fixtures/same place as others at some point so its more obvious. Thanks.

cykoder (approved):
lgtm but needs package.json version bump?

PR #442, opened 2024-07-26
Fix getDidMethodKeyDetail NoDIDError invalid DID string

cykoder:


olegnn (approved):

olegnn (commented):
I guess that several more errors like this exist in this file.

cykoder:
> I guess that several more errors like this exist in this file.

This is the only one with `did:key:dock` and the others are like ``did:dock:${didIdentifier}`` which I think would be mostly correct, didIdentifier seems to be hex?

PR #446, opened 2024-08-23
Revert "Revert anoncreds upgrade"

cykoder:
Reverts docknetwork/sdk#445

Needs investigating why the older anoncreds (KVAC) credentials failed to verify after this change

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@docknetwork/sdk",
-  "version": "9.2.0",
+  "version": "9.3.0",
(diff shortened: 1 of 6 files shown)


cykoder:
Actually it seems that the SDK didnt have a bug, but the wallet app's older SDK version couldn't do proper presentation cryptography (produces invalid result) due to the SDK changes here. This caused a submitted presentation to be invalid, this update should be co-ordinated with wallets as its technically a breaking change

maycon-mello (approved):

PR #463, opened 2024-10-23
Update README.md

esplinr:
Corrections, clarifications, and an additional deprecation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,23 +4,27 @@
 
 The Dock SDK is an opensource library that powers Dock's SaaS API. It provides the credential management, cryptography, and blockchain storage features for Dock's supported credential variants:
 
-* W3C VC JSON-LD ed25519 credential using StatusList 2021 for revocation targetted at interoperability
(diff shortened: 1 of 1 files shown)


esplinr:
Is anything inaccurate?

Are there any other deprecations we should include?

mike-parkhill (approved):

olegnn:
I’m not sure why `akworks` is listed separately, as it’s not a direct dependency of the SDK. Instead, it’s a dependency of the [Dock crypto library](https://github.com/docknetwork/crypto), which is itself a dependency of the [Dock WASM crypto library](https://github.com/docknetwork/crypto-wasm). This, in turn, is a dependency of the [Dock TS crypto wrapper library](https://github.com/docknetwork/crypto-wasm-ts), which ultimately becomes a dependency of the SDK.
Also, this library includes full anoncreds support using Pointcheval-Sanders and BBS+ schemes.

cykoder (approved):

PR #469, opened 2024-11-05
Fix exports

olegnn:


examples/schema-validation.js
```diff
@@ -1,7 +1,7 @@
 import { Schema } from '@docknetwork/credential-sdk/modules/abstract/schema';
 
 // BOL Schema, its valid
-import bolSchema from './schemas/bol';
+import bolSchema from './schemas/bol.js';
```
cykoder, line 4: why the need to add .js here?

cykoder (approved):

PR #487, opened 2024-11-28
`StatusListCredential` module for `cheqd`

olegnn:


packages/credential-sdk/src/types/did/document/verification-method copy.js
```diff
@@ -0,0 +1,186 @@
+import {
```
cykoder, line 1: was this file intended to be committed?

cykoder (approved):
lgtm aside from failing CI and one file that looks like it was included accidentally, as its named "copy"

cykoder (approved):

PR #502, opened 2025-01-13
Align with the latest `cheqd` node version (`3.1.1`)

olegnn:


packages/credential-sdk/src/utils/misc.js
```diff
@@ -1,67 +1,7 @@
 /* eslint-disable max-classes-per-file */
-import elliptic from 'elliptic';
-
-import { sha256 } from 'js-sha256';
 import { PatternMatcher } from './generic';
 import { METHOD_REG_EXP_PATTERN } from '../resolver/generic/const';
 
-const EC = elliptic.ec;
-const secp256k1Curve = new EC('secp256k1');
```
cykoder, line 9: any removed imports from this file that arent used anywhere else should be removed from package.json too (And yarn.lock updated ofc). was that checked?

olegnn: All of them are utilized by the [`Secp256k1Keypair`](https://github.com/docknetwork/sdk/blob/master/packages/credential-sdk/src/keypairs/keypair-secp256k1.js).

cykoder: got it, thought it was removing of logic not just cleaning up from when it was moved in #449

cykoder (dismissed):

cykoder (approved):

PR #528, opened 2025-02-21
Use cjs Cheqd SDK

cykoder:
Temporarily revert https://github.com/docknetwork/sdk/pull/525 as it has issues with package resolution


olegnn (approved):

olegnn (changes requested):
Can you update `yarn.lock`?

olegnn (approved):

PR #542, opened 2025-09-25
Support various key identifiers

cykoder:
Allows to use key-1, keys-1 etc


cykoder:
the existing tests cover the logic (if #keys-1 works here, #key-1 will too due to regex rules), and there are more tests in https://github.com/docknetwork/sdk/pull/543 for document parsing with different identifier formats. i had to include that in another PR because of the other changes needed to get that document working

PR #585, opened 2026-04-13
Custom did:key implementation for anoncreds

cykoder:


gasher (approved):

mike-parkhill (approved):
Should we push an update to the universal resolver repo for these?

maycon-mello (approved):

cykoder:
> Should we push an update to the universal resolver repo for these?

I don't think itd be accepted as we arent standardised, we'd have to update the existing universal resolver did:key implementation

PR #607, opened 2026-07-18
Create crypto-utils package

cykoder:

PR #608, opened 2026-07-18
AP2 package with receipt issuance and verification

cykoder:
Relies on https://github.com/docknetwork/sdk/pull/607

### DIFF
--- a/packages/ap2/README.md
+++ b/packages/ap2/README.md
@@ -1,0 +1,4 @@
+# @docknetwork/ap2
+
+TODO: write this
+
--- a/packages/ap2/README.md
+++ b/packages/ap2/README.md
@@ -1,4 +1,109 @@
 # @docknetwork/ap2
 
-TODO: write this
+AP2 receipt issuance and verification using the lightweight Dock crypto
+utilities. Receipts are compact ES256 JWTs compatible with the AP2 receipt
+format.
 
+```js
+import {
+  buildPaymentReceipt,
+  computeMandateReference,
+  signReceipt,
+  verifyPaymentReceipt,
+} from '@docknetwork/ap2';
+import { Secp256r1Keypair } from '@docknetwork/crypto-utils/keypairs';
+
+const keypair = Secp256r1Keypair.random();
+// Use the exact compact presentation received from the Shopping Agent.
+const closedMandatePresentation = receivedClosedMandatePresentation;
+const receipt = buildPaymentReceipt({
+  status: 'Success',
+  iss: 'mpp.acme',
+  iat: Math.floor(Date.now() / 1000),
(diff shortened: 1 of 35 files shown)


packages/ap2/README.md
```diff
@@ -0,0 +1,109 @@
+# @docknetwork/ap2
+
+AP2 receipt issuance and verification using the lightweight Dock crypto
+utilities. Receipts are compact ES256 JWTs compatible with the AP2 receipt
```
mike-parkhill: Should we link off to AP2 documentation, or at least expand on what AP2 is?

packages/crypto-utils/src/vc/sd-jwt.js
```diff
@@ -0,0 +1,138 @@
...
+ *
+ * @param {string} presentation
+ * @returns {{issuerJwt: string, disclosures: string[], kbJwt?: string}}
+ */
+export function parseSdJwtPresentation(presentation) {
+  if (typeof presentation !== 'string' || presentation.length === 0) {
+    throw new TypeError('SD-JWT presentation must be a non-empty string');
+  }
+  if (!presentation.includes('~')) {
+    throw new Error('SD-JWT presentation must contain a tilde');
```
mike-parkhill, line 43: Is that part of the standard?

packages/ap2/src/utils.js
```diff
@@ -0,0 +1,91 @@
...
+  if (!validate) {
+    throw new Error(`Unsupported receipt type: ${type}`);
+  }
+  if (!validate(receipt)) {
+    throw new TypeError(
+      `Invalid ${type} receipt: ${formatValidationErrors(validate.errors)}`,
+    );
+  }
+
+  return { ...receipt };
```
mike-parkhill, line 46: any risk due to this being a shallow clone rather than a full clone?

mike-parkhill (commented):
From Claude:
The README is upfront that mandatePresentation verification only checks the reference hash, not the mandate's own signature/delegation/audience/etc. — good, this avoids a false sense of security, but it does mean misuse (skipping the separate mandate-verification step) is easy for an integrator to get wrong. Consider whether verifyReceipt should optionally accept a mandate-verification callback/result rather than relying entirely on documentation to prevent this footgun

packages/ap2/src/utils.js
cykoder: standard claims are all primitives (status, iss, iat, reference, ids, error strings) so { ...receipt } is the same as a deep clone

packages/crypto-utils/src/vc/sd-jwt.js
cykoder: yeah, sd-jwts are tilde separated

mike-parkhill (approved):

PR #616, opened 2026-07-31
Ap2/verify cnf and reference binding

mike-parkhill:
Harden AP2 Closed Mandate verification — close every gap where trust was silently deferred to the caller instead of enforced.

verifyClosedCheckoutMandate/verifyClosedPaymentMandate accumulated several places where a signature check passed but the thing that actually mattered wasn't verified. Since this package has no production usage yet, each gap is closed outright rather than kept as a fallback:

Caller-supplied-key fallback removed. publicKey/holderJwk options are gone; the verification key is always derived from the Open Mandate's own cnf.jwk. openMandatePresentation is now required.
Open Mandate issuer now verified. New required userPublicKey param — the caller's independently-resolved trusted key for the delegating User — checked against the Open Mandate's own envelope. Without this, an attacker could fabricate an entire Open + Closed Mandate chain naming their own key in cnf.jwk, and every other check would pass. resolveOpenCheckoutMandateContent/resolveOpenPaymentMandateContent gained a matching optional userPublicKey for direct callers.
Envelope typ now checked (kb+sd-jwt), rejecting an Open Mandate or other envelope type fed in where a Closed one is expected.
Replay protection: nonce is now enforced. New required expectedNonce param, checked against the envelope's nonce. Closed Mandate envelopes carry no exp of their own, so previously a validly-signed presentation could be replayed indefinitely — nonce was accepted at signing time but never checked at verification time.
Also:

verifyClosedPaymentMandate gained openCheckoutMandatePresentation (verifies the Open Payment Mandate's payment.reference binding, previously left to the caller) and resolveOpenCheckoutMandateContent was added (mirrors the payment-side resolver).
Docs: documented that AP2 v0.2 defines no revocation mechanism at all (confirmed against the spec — explicitly out of scope, exp is the only verifier-enforceable lifecycle control and is only RECOMMENDED, not required); added an example resolving userPublicKey from a DID.
Added .d.ts type declarations to this package and @docknetwork/crypto-utils (unrelated tooling fix, bundled in since both landed in this window).
Breaking: yes, but pre-1.0 and unreleased — no downstream consumers exist yet. Confirmed via repo-wide search across wallet-sdk (no usage), ap2-example-app (no direct usage, only via MCP calls to truvera-mcp-server), and truvera-mcp-server (3 call sites in wallet-server/truvera-api need updating — tracked separately, not part of this PR).


maycon-mello (dismissed):
looks good

packages/ap2/src/mandates.js
```diff
@@ -575,18 +697,35 @@ export function verifyClosedPaymentMandate(
...
 
     return {
       verified: true,
       content: validatedContent,
       protectedHeader,
       ...(transactionIdVerified === undefined ? {} : { transactionIdVerified }),
-      ...(sdHashVerified === undefined ? {} : { sdHashVerified }),
+      ...(referenceVerified === undefined ? {} : { referenceVerified }),
+      sdHashVerified: true,
+      openMandateIssuerVerified: true,
```
dockscp, line 749: I see multiple default `true` assignments here to verification fields - which makes me nervous. Can you perhaps double-check it's ok?
Is it ok that `referenceVerifier` can be false but `verified` is true? The [respective test case](https://github.com/docknetwork/sdk/blob/17c689e32c23aa5526c7501c7939a6bd60a0fd65/packages/ap2/tests/mandates.test.js#L852-L853) seems to assert this is the expected behavior, but just wanted to make sure?

packages/ap2/src/mandates.js
```diff
@@ -526,46 +623,71 @@ export function verifyClosedCheckoutMandate(
...
     const { issuerJwt, disclosures } = parseSdJwtPresentation(presentation);
-    const result = verifyJWT(issuerJwt, resolvePublicKey({ publicKey, holderJwk }), { algorithms: ['ES256'] });
+
+    const openPaymentContent = resolveOpenPaymentMandateContent(
+      openMandatePresentation,
+      { userPublicKey, currentDate, clockTolerance },
+    ).content;
+    const verificationKey = jwkToSecp256r1PublicKey(openPaymentContent.cnf.jwk);
+
+    const result = verifyJWT(issuerJwt, verificationKey, { algorithms: ['ES256'] });
```
dockscp, line 693: nitpicking: I understand that we only support P-256 keys but this static assignment might bite us in the future 😅

dockscp (commented):
nitpick: Just because I happened to see it -> https://github.com/docknetwork/sdk/blob/17c689e32c23aa5526c7501c7939a6bd60a0fd65/packages/ap2/src/mandates.js#L417-L425

Broadly speaking, we should also check `iat`, not only `exp`.

dockscp (commented):
https://github.com/docknetwork/sdk/blob/17c689e32c23aa5526c7501c7939a6bd60a0fd65/packages/ap2/src/receipts.js#L140

Related to my first comment, can you double check if this is ok?

packages/ap2/src/mandates.js
mike-parkhill: Fixed in latest commit

packages/ap2/src/mandates.js
mike-parkhill: Yeah, but it's all that AP2 demands right now

mike-parkhill:
@dockscp I think all your concerns are addressed now.

dockscp (approved):

PR #617, opened 2026-08-05
extract JsonWebKey from @transmute/json-web-signature

maycon-mello:
- New file src/vc/crypto/JsonWebKey.js — extracted JsonWebKey from @transmute/json-web-signature (you only ever imported JsonWebKey, which never touches jsonld/http-client). Repointed the 2 imports in JsonWebSignature2020.js and vc/helpers.js.
  
  - package.json: removed @transmute/json-web-signature; added its 4 clean sub-deps (ed25519-key-pair, secp256k1-key-pair, jose-ld, web-crypto-key-pair) as direct deps; bumped jsonld-signatures ^9.3.1 → ^10.0.0 (uses jsonld@6, same major you already run);

### DIFF
--- a/packages/credential-sdk/src/vc/crypto/JsonWebKey.js
+++ b/packages/credential-sdk/src/vc/crypto/JsonWebKey.js
@@ -1,19 +1,23 @@
-import {
-  Ed25519KeyPair,
-} from '@transmute/ed25519-key-pair';
-import {
-  Secp256k1KeyPair,
-} from '@transmute/secp256k1-key-pair';
-
+// Local copy of JsonWebKey from @transmute/json-web-signature@0.7.0-unstable.82,
+// extracted to drop that package's broken jsonld@5 -> @digitalbazaar/http-client@1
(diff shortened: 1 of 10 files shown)


mike-parkhill (changes requested):
Some linting errors and I think there's still a Copilot comment that's unresolved

packages/credential-sdk/src/vc/crypto/JsonWebKey.js
```diff
@@ -0,0 +1,157 @@
...
+      options.secureRandom = () => crypto.randomBytes(32);
+    }
+    const kp = await KeyPair.generate({
+      kty: options.kty,
+      crvOrSize: options.crv,
+      secureRandom: options.secureRandom,
+    });
+    const { detached } = options;
+    return useJwa(kp, { detached });
+  };
```
cpatsonakis, line 142: perhaps we should not mutate the caller-provided options object here? We can use a local `secureRandom` variable here I believe.

mike-parkhill (approved):

